PR #65, opened 2019-08-12
Move README to root

hausdorff:


README.md
```diff
@@ -1,3 +1,119 @@
...
+### Install the dev version of the CLI
+
+```sh
+# NOTE: after this step, you'll need to make sure the dev version of the Pulumi
+# CLI comes before your system install. If you're on macOS, usually you need to
+# make sure that `$(go env GOBIN)` comes before `/usr/local/bin` on your $PATH.
+git clone git@github.com:pulumi/pulumi.git
+cd pulumi
+make ensure build install
+p version # should be > v1.0.0-beta.1
```
clstokes:
```suggestion
pulumi version # should be > v1.0.0-beta.1
```
Use the full command name.

clstokes:
Can we also add:
- explanation of enforcement level
- explanation of differing behavior during preview and update

README.md
```diff
@@ -1,3 +1,80 @@
...
 # Pulumi policy framework
+
+Status: **beta release.**
+
+Define and manage policy for cloud resources deployed through for Pulumi.
+
+Policy rules run during `pulumi preview` and `pulumi up`, asserting that cloud resource definitions
+comply with the policy immediately before they are created or updated.
+
+During `preview`, all every rule is run on every resource, and policy violations are batched up into
```
clstokes:
```suggestion
During `preview`, every rule is run on every resource, and policy violations are batched up into
```

PR #77, opened 2019-08-21
Throw when a field is unknown

### LINKED ISSUE #63: Unknown values are not being stripped from analyze calls
They should be undefined, but they get passed through as the string: `04da6b54-80e4-46f7-96ec-b56ff0331ba9`

hausdorff:
When a policy is run during preview, it may access properties that can't
be known until update. This commit will cause unknown property accesses
to result in throwing an `UnknownValueError`. This allows the policy
plugin to catch unknown value accesses and report to the user that the
policy can't complete during preview, and hence may fail at runtime.

Fixes #63.

## Demo

Suppose we have the following policy:

```typescript
        {
            name: "kube-stuff",
            description: "kube services",
            rules: [
                typedRule(k8s.core.v1.Service.isInstance, it =>
                    assert.strictEqual(it.metadata.labels["foo"], "bar"),
                ),
            ],
        },
```

and the following app code, with computed value in the labels:

```typescript
const nginx = new k8s.apps.v1.Deployment(appName, { [...] });

const frontend = new k8s.core.v1.Service(appName, {
    metadata: { labels: { foo: nginx.status.readyReplicas.apply(r => r.toString()) } },
    spec: {
        type: "LoadBalancer",
        ports: [{ port: 80, targetPort: 80, protocol: "TCP" }],
        selector: appLabels,
    },
});

```

running this during preview will result in: (edit: it also has some testing changes from pulumi/pulumi—oops!)

unknown

### DIFF
--- a/sdk/nodejs/policy/proxy.ts
+++ b/sdk/nodejs/policy/proxy.ts
@@ -1,0 +1,130 @@
+// Copyright 2016-2019, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * `unknownCheckingProxy` takes a set of resource inputs, and returns a wrapped version that
+ * intercepts all property accesses to check if they are unknown, throwing an `UnknownValueError` if
+ * so. For example, if the resource inputs contains a reference to an IP address that can't be known
+ * until the resource has been initialized by the cloud provider, this value would be unknown during
+ * `pulumi preview`, and only filled in during the update.
+ *
+ * @param toProxy resource inputs to create a proxy for
+ */
+export function unknownCheckingProxy<T>(toProxy: any): any {
+    return proxyHelper(toProxy, []);
+}
+
+/**
+ * `proxyHelper` is a helper for the `unknownCheckingProxy` function.
+ * @param toProxy resource inputs to create a proxy for
+ * @param propsAcc accumulates the property path, e.g., for `resc.foo.bar`, this would be `["foo",
+ * "bar"]`
+ */
+function proxyHelper<T>(toProxy: any, propsAcc: (keyof T)[]): any {
+    if (!(toProxy instanceof Object)) {
+        return toProxy;
+    }
+
+    return new Proxy(toProxy, {
+        get: (obj, prop: keyof T) => {
+            const newProps = propsAcc.concat([prop]);
+            const field = obj[prop];
+            if (isUnknown(field)) {
+                throw new UnknownValueError<T>(field, newProps);
+            }
+            return proxyHelper(field, newProps);
+        },
+    });
+}
+
+/**
+ * UnknownValueError indicates that a property we attempted to access in some cloud resource is
+ * unknown. For example, during preview, some resource fields (such as an allocated IP address)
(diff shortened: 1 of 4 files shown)


sdk/nodejs/policy/proxy.ts
```diff
@@ -0,0 +1,106 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export function proxy<T>(toProxy: any): any {
+    return proxyHelper(toProxy, []);
+}
+
+function proxyHelper<T>(toProxy: any, props: (keyof T)[]): any {
```
CyrusNajmabadi: i'm pretty sure i'm understanding it.  but it prob wouldn't hurt to document this.  i.e. this array represents the sequence of property names accessed off the object (at least that's how i understand it).

sdk/nodejs/policy/proxy.ts
```diff
@@ -0,0 +1,106 @@
...
+        },
+    });
+}
+
+export class UnknownValueError<T> extends Error {
+    constructor(public readonly unknownType: string, public readonly props: (keyof T)[]) {
+        super(
+            `${unknownToString(unknownType)} value at .${props.join(
+                ".",
+            )} can't be statically computed`,
```
CyrusNajmabadi: nit: just make this a one-liner :)

sdk/nodejs/policy/proxy.ts
```diff
@@ -0,0 +1,106 @@
...
+            if (isUnknown(field)) {
+                throw new UnknownValueError<T>(field, newProps);
+            }
+            return proxyHelper(field, newProps);
+        },
+    });
+}
+
+export class UnknownValueError<T> extends Error {
+    constructor(public readonly unknownType: string, public readonly props: (keyof T)[]) {
```
CyrusNajmabadi: query: does anyone uses the args outside of this error?  If not, consider not making them properties.

sdk/nodejs/policy/proxy.ts
```diff
@@ -0,0 +1,106 @@
...
+        super(
+            `${unknownToString(unknownType)} value at .${props.join(
+                ".",
+            )} can't be statically computed`,
+        );
+    }
+}
+
+function isUnknown(o: any): any {
```
CyrusNajmabadi: `boolean` return type instead of `any`.

sdk/nodejs/policy/proxy.ts
CyrusNajmabadi: ah, i see you use them for testing purposes.

sdk/nodejs/policy/proxy.ts
```diff
@@ -0,0 +1,106 @@
...
+            const field = obj[prop];
+            if (isUnknown(field)) {
+                throw new UnknownValueError<T>(field, newProps);
+            }
+            return proxyHelper(field, newProps);
+        },
+    });
+}
+
+export class UnknownValueError<T> extends Error {
```
CyrusNajmabadi: have a short doc for an exported class.

sdk/nodejs/policy/proxy.ts
hausdorff: I defer all style issues to the formatter—if this is _really_ important I can maybe shorten one of the variables.

hausdorff: Let's pull some of those computations out into local variables.

sdk/nodejs/policy/proxy.ts
hausdorff: I don't think I know what this means, but shortening the call to super means that we have to have local variables before `super`, which means that we need to set these properties manually vs having them baked into the constructor.

CyrusNajmabadi: so, `constructor(public whatever)` means you're declaring both a parameter and a property on hte class.  I was asking if you actually needed the property or not.  I see you actually use the property later on in tests, so this is ok.

PR #97, opened 2019-08-29
Ignore/remove package-lock.json for examples

clstokes:


clstokes:
> @clstokes I'm told if we do this we should also set the version to `^0.1.0` in package.json. Would you mind doing that too?

Done.

justinvp:
Going to close since we've moved the examples to the examples repo.

PR #115, opened 2019-09-27
Add support for secret values in policies

### LINKED ISSUE #30: Decide how to handle secrets in `Analyze`
Following up from https://github.com/pulumi/pulumi/pull/2955, we should decide explicitly how we want to handle secrets. This PR makes no attempt to do this.

hausdorff:
Fixes #30.

### DIFF
--- a/sdk/nodejs/policy/deserialize.ts
+++ b/sdk/nodejs/policy/deserialize.ts
@@ -58,8 +58,7 @@ function deserializeProperty(prop: any): any {
     } else if (prop instanceof Array) {
         const elems: any[] = [];
         for (const e of prop) {
-            prop = deserializeProperty(e);
-            elems.push(unwrapRpcSecret(prop));
+            elems.push(deserializeProperty(e));
         }
 
         return elems;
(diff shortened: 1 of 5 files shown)


pgavlin:
What's the risk that a user unintentionally exposes a secret (e.g. by leaking it into logs or other persisted data) from a policy?

hausdorff:
@pgavlin I'm actually not that concerned with leaking secrets to the Pulumi service through logs. I'm a little more concerned that, because policies have to operate on raw, it's so easy to leak secrets generally—printing, or even reporting the policy violation, and so on.

There don't seem to be any really great answers here, and when I looked into it, ~all policy systems effectively have this issue. The alternatives are:

* Simply don't run policies if they touch secret data, in the same way we don't for unknowns. I strongly believe this won't fly with customers—they want a policy system to disallow _any_ resource in violation of a policy, even if it's secret.
* Wrap all secrets in a proxy object that proxies `toString` -> `[secret]`. This is pretty invasive, and means a lot of things that depend on implicit `toString`, (_e.g._, string interpolation) will now subtly not do what users want. It also means that you won't have the opportunity to print a useful error message, which itself is probably not a huge deal, since you have access to the code anyway.

Thoughts?

pgavlin:
> policies have to operate on raw, it's so easy to leak secrets generally—printing, or even reporting the policy violation, and so on.

FWIW, Pulumi programs have the same issue: secret values are available in plaintext to the user's program inside an `apply`. As long as we don't have additional avenues for persistence (e.g. the state file, which is solved for in Pulumi programs using `Output<>`'s propagation of secretness), I think we're fine as-is.

sdk/nodejs/policy/deserialize.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+function isRpcSecret(obj: any): boolean {
+    return obj && obj[specialSigKey] === specialSecretSig;
+}
+
+/**
+ * unwrapRpcSecret returns the underlying value for a secret, or the value itself if it was not a
+ * secret.
+ */
+function unwrapRpcSecret(obj: any): any {
```
pgavlin: This shouldn't be necessary, right? We're never returning a first-class secret from `deserializeProperty`, so all of them should already be unwrapped.

hausdorff:
That's true, the situation is a little worse here though—think about if a policy violation occurs on a password or something. That password will probably emitted as a policy violation event.

That said, generally speaking, a policy _rejects_ a value, so even if it's secret, you'll have to change it anyway. So in practice I suspect this will not be as big of a deal as it seems.

sdk/nodejs/policy/deserialize.ts
hausdorff: We can get rid of both `isRpcSecret` and `unwrapRpcSecret`.

PR #124, opened 2019-10-10
Cleanup comments

chrsmith:
Add some comments to important things that were missing them, and correct comments that were no longer accurate. e.g. "A function that returns true if a resource definition violates some policy." should now be "A function that throws an `AssertionError` if a resource definition violates some policy."

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -35,11 +35,8 @@ export class PolicyPack {
     constructor(private name: string, args: PolicyPackArgs) {
         this.policies = args.policies;
 
-        //
         // TODO: Wire up version information obtained from the service.
-        //
         const version = "1";
-
         serve(this.name, version, this.policies);
     }
 }
@@ -55,7 +52,7 @@ export type EnforcementLevel = "advisory" | "mandatory";
  * is violated.
  */
 export interface Policy {
-    /** An ID for the policy. Must be unique to the current policy set. */
+    /** An ID for the policy. Must be unique within the current policy set. */
     name: string;
 
     /**
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -48,8 +48,19 @@ import { version } from "./version";
 
 // ------------------------------------------------------------------------------------------------
 
+// Flag indicating whether or not a gRPC service is currently running for this process.
 let serving = false;
 
+/**
+ * Starts the gRPC server to communicate with the Pulumi CLI client for analyzing resources.
+ *
+ * Only one gRPC server can be running at a time, and the port the server is running on will
+ * be written to STDOUT.
+ *
+ * @param policyPackName Friendly name of the policy pack.
+ * @param policyPackVersion Version of the policy pack SDK used.
+ * @param policies The policies to be served.
+ */
 export function serve(policyPackName: string, policyPackVersion: string, policies: Policies): void {
     if (serving !== false) {
         throw Error("Only one policy gRPC server can run per process");


sdk/nodejs/policy/policy.ts
```diff
@@ -13,37 +13,76 @@
...
-import { Resource } from "@pulumi/pulumi";
 import * as q from "@pulumi/pulumi/queryable";
+
 import { serve } from "./server";
 
+/**
+ * Arguments for constructing a new PolicyPack object.
+ */
 export interface PolicyPackArgs {
+    // Policies to associate with the policy pack.
```
justinvp: Let's make this a `/** */` doc comment style comment, since this is a library. That way, the comment will show up in intellisense when using the API in IDEs like VS Code.

sdk/nodejs/policy/policy.ts
```diff
@@ -13,37 +13,76 @@
...
 }
 
-/** A function that returns true if a resource definition violates some policy. */
+/**
+ * A Rule verifies the resource provided matches some expected state, and if not
+ * throws an exception of type AssertionError with a description of the problem.
+ * Otherwise, it is assumed the resource has been successfully verified.
+ *
+ * For example:
+ * ```
```
justinvp: Add a blank line above the code fence and specify a language, so it'll render correctly when we generate the API docs with proper syntax highlighting.

```suggestion

 * ```typescript
```

sdk/nodejs/policy/policy.ts
```diff
@@ -13,37 +13,76 @@
...
+ * };
+ * ```
+ */
 export type Rule = (type: string, properties: any) => void;
 
+/**
+ * typedRule provides a convenient shorthand for creating new rules that are only
+ * executed against resources of a specific type. For example:
+ *
+ * ```
```
justinvp:
```suggestion
 * ```typescript
```

sdk/nodejs/policy/policy.ts
chrsmith: Done. I assumed it would "just work", but even then we should be consistent.

sdk/nodejs/policy/policy.ts
```diff
@@ -13,37 +13,77 @@
...
+ * import * as assert from "assert";
+ * import * as aws as "@pulumi/aws";
+ *
+ * typedRule(aws.s3.Bucket.isInstance, (s3Bucket: any) => {
+ *     assert.ok(it.acl === "public-read", `The S3 bucket s3://${s3Bucket["name"]} is public.`);
+ * });
+ * ```
+ *
+ * @param filter is a type guard, verifying the provided value is an instance of the
+ * type the resource the policy whishes to inspect. Typically Pulumi libraries have
```
ekrengel: wishes

sdk/nodejs/policy/server.ts
```diff
@@ -39,8 +40,19 @@ import { version } from "./version";
 
 // ------------------------------------------------------------------------------------------------
 
+// Flag indicating whether or not a gRPC service is currently running for this process.
 let serving = false;
 
+/**
+ * Starts the gRPC server to communication with the Pulumi CLI client for analyzing resources.
```
ekrengel: to communicate

sdk/nodejs/policy/policy.ts
chrsmith: Done

sdk/nodejs/policy/server.ts
chrsmith: Done.

justinvp:
I am going to fold these changes into a change I have in progress to fix-up `@pulumi/policy` for the docs, which makes some tweaks to doc comments.

Issue #120, opened 2019-10-07
Improve the Readme


ekrengel:
Walk through the readme and make sure its 💯

PR #126, opened 2019-10-14
Improve readme + move examples

ekrengel:
* Move examples to `examples` repo
* Improve readme

Fixes https://github.com/pulumi/pulumi-policy/issues/120

(At least until we make more changes to the SDK :) )


justinvp:
Would it be worth keeping some examples in this repo? Most of our other libraries have a set of examples in their repo (e.g. see https://github.com/pulumi/pulumi-aws) in addition to examples in the examples repo, and I believe such examples are used as tests. I'm not sure what we'd keep here vs. have in the examples repo, though.

ekrengel:
> Would it be worth keeping some examples in this repo? 

I think so, but I also think youre right that its a hard call what to keep here vs move. I am going to opt for moving everything for now. We can always add back examples when we add tests if needed.

Issue #87, opened 2019-08-28
API feedback


joeduffy:
Here are a few bits of API feedback. I'll add more as I go:

- It's odd that `rules` can have a single rule, since the property name is plural. Any reason not to just have a `rule` property for the singular case (which I assume will be most common)?

- `enforcementLevel` is a mouthful. Furthermore, I find it odd to define enforcement level on the policy rule _definition_, rather than its _application_. Should we just call this `severity`?

- I find the `typedRule` construct a bit awkward to interact with. Instead of

    ```typescript
    rule: typedRule(aws.s3.Bucket.isInstance, bucket => {}),
    ```

    is there any way it could be something like this instead?

    ```typescript
    rule: new Rule<aws.s3.Bucket>(bucket => {}),
    ```

hausdorff:
```typescript
rule: new Rule<aws.s3.Bucket>(bucket => {}),
```

How would this work? You need the `isInstance` to perform the filter objects, right? You might be able to do something like (this is pseudocode) `Rule<T implements { isInstance: (o: any) => o is T }>` but I have not tried this to see if it would work.

hausdorff:
> It's odd that `rules` can have a single rule, since the property name is plural. Any reason not to just have a `rule` property for the singular case (which I assume will be most common)?

I am not opposed to this if everyone thinks this is more user-friendly, but I think it's weird to have two fields for rules, and I don't know what the behavior should be if you set both `rules` and `rule`.

hausdorff:
> `enforcementLevel` is a mouthful. Furthermore, I find it odd to define enforcement level on the policy rule _definition_, rather than its _application_. Should we just call this `severity`?

We put this on the `Policy` itself because (1) it's more annoying to have to set them all in the console, and (2) other systems include it in the rule definition—but, eventually, when we have config values, we will make this overridable.

As for me, I'm not sure what `severity` means if it's not an enforcement level. Again—whatever people find more intuitive, but I personally prefer `enforcementLevel`.

also cc @clstokes and @ekrengel on all of the above.

joeduffy:
> I think it's weird to have two fields for rules

Personally, I think it's weirder to have a property named `rules` that accepts a single rule!

> I'm not sure what severity means if it's not an enforcement level. Again—whatever people find more intuitive, but I personally prefer enforcementLevel.

[...]

Issue #78, opened 2019-08-22
Expose resource provider configuration for policies


clstokes:
Policy users will want to develop policies to check a resource's provider - e.g. `require this resource to be in us-east-1`. We should expose the provider information from the resource to enable this.

joeduffy:
More generally, it seems we should enable the full programming model, with a few exceptions.

These should be legal

- Current project/stack names
- Provider configuration
- Resource provider data sources
- StackReferences

While these likely should not

- Allocating new resources
- Exporting stack outputs

joeduffy:
Attempting to use `getProject` inside of a policy leads to an error:

```
TypeError: pulumi.getProject is not a function
    at /Users/joeduffy/temp/iam-analyzer-policy/index.ts:22:44
    at Generator.next (<anonymous>)
    at /Users/joeduffy/temp/iam-analyzer-policy/index.ts:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/joeduffy/temp/iam-analyzer-policy/index.ts:4:12)
    at Object.validateStack (/Users/joeduffy/temp/iam-analyzer-policy/index.ts:12:108)
    at Object.<anonymous> (/Users/joeduffy/temp/iam-analyzer-policy/node_modules/@pulumi/policy/server.js:204:49)
    at Generator.next (<anonymous>)
    at /Users/joeduffy/temp/iam-analyzer-policy/node_modules/@pulumi/policy/server.js:21:71
```

justinvp:
Related to #42

justinvp:
> `pulumi.getProject is not a function`

Even when we make the necessary data available to the policy pack program, the reason it's saying `getProject` is not a function, is because inside `@pulumi/pulumi`, when we import the `./metadata` module in `index.ts` to export the `getProject` and `getStack` functions, the `/.metadata` module returned is empty. This is indicative of circular references in our modules (node is returning an empty module to avoid infinite loop). It doesn't happen when accessed from a Pulumi program, but is happening when accessed from a Policy Pack program. Will need to unravel some (ideally all) of our circular dependencies to make this work.

Using the [`madge`](https://www.npmjs.com/package/madge) tool inside the `sdk/nodejs` dir in `pulumi/pulumi` repo yields:

```sh
$ madge -c .
Processed 179 files (1.6s) (6 warnings)

✖ Found 16 circular dependencies!

[...]

PR #131, opened 2019-10-25
API changes

justinvp:
Apologies for the long description.  This PR implements the latest API proposal, addressing some feedback [since we discussed earlier this week](https://docs.google.com/document/d/1b-8GKeL4zkXVn-BjiO-H0BDeY6hgg-lxioY3psDG1zs/edit#). Please read it over and take a look at the changes in this PR and the associated PR that updates the examples https://github.com/pulumi/examples/pull/441

Some changes since we discussed earlier this week:

 - This demonstrates a `validateTypedResource` helper for stronger typed resource checking, similar to the previous `typedRule` helper.

    Here's an example:

    ```typescript
    new PolicyPack("policy-pack-typescript", {
        policies: [{
            name: "s3-no-public-read",
            description: "Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
            validateResource: validateTypedResource(aws.s3.Bucket.isInstance, (it, args, reportViolation) => {
                if (it.acl === "public-read" || it.acl === "public-read-write") {
                    reportViolation(
                        "You cannot set public-read or public-read-write on an S3 bucket. " +
                        "Read more about ACLs here: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html");
                }
            }),
        }],
    });
    ```

 - I keep waffling on changing `EnforcementLevel`, `"advisory"`, and `"mandatory"` to `Severity`, `"warning"`, and `"error"`. I don't feel strongly about it, but was initially leaning towards the less verbose `Severity`. However, Cameron pointed out we might want to add a `"softMandatory"` or `"requiresApproval"` value in the future, which doesn't work as well with `Severity`, so I've left it `EnforcementLevel` for now. Still open to changing.

 - Originally, I specified the `Policy` as:

    ```typescript
    export interface Policy {
        name: string;
        description: string;
        enforcementLevel: EnforcementLevel;

        validateResource?: ResourceValidation;
        validateStack?: StackValidation;
    }
    ```

    Which would allow setting either `validateResource` or `validateStack` on a policy. However, [Cameron's feedback](https://docs.google.com/a/pulumi.com/document/d/1b-8GKeL4zkXVn-BjiO-H0BDeY6hgg-lxioY3psDG1zs/edit?disco=AAAADempQdU) was that we should not allow both.

    To address, I've separated these out as follows so that the compiler will enforce defining either a `validateResource` or `validatStack`:

    ```typescript
    export interface Policy {
        name: string;
        description: string;
        enforcementLevel: EnforcementLevel;
    }

    export interface ResourceValidationPolicy extends Policy {
        validateResource: ResourceValidation | ResourceValidation[];
    }

    export interface StackValidationPolicy extends Policy {
        validateStack: StackValidation;
    }

    export type Policies = (ResourceValidationPolicy | StackValidationPolicy)[];
    ```

    If the additional types add too much cognitive load, we could consider going back to the simpler design with two optional properties on the main `Policy` interface. We could also consider making it a runtime check to enforce only setting one of the two properties (if we felt strongly about enforcing this). Or, we could allow both to be set, but generally recommend setting one or the other.

 - You'll probably notice that I changed:

    ```typescript
    validateResource: ResourceValidation;
    ```

    to:

    ```typescript
    validateResource: ResourceValidation | ResourceValidation[];
    ```

    I know we discussed how it didn't make a lot of sense to have `rules` which accepted a single `rule` (despite using this pattern elsewhere like in AWSX), and how the assumption is that most polices will have a single rule. When updating the examples, I found it was just easier to be able to provide multiple validation functions to be called for certain policies that are checking similar things across multiple resource types (e.g. EC2 `Instance`, `LaunchConfiguration`, or `LaunchTemplate` are [using only approved AMIs](https://github.com/pulumi/examples/pull/441/files#diff-02168f5f7813feed7b3872ba18a781feR33)).

    ```typescript
    {
        name: name,
        description: "Instances should use approved AMIs.",
        validateResource: [
            validateTypedResource(aws.ec2.Instance.isInstance, (it, args, reportViolation) => {
                if (amis && !amis.has(it.ami)) {
                    reportViolation("EC2 Instances should use approved AMIs.");
                }
            }),
            validateTypedResource(aws.ec2.LaunchConfiguration.isInstance, (it, args, reportViolation) => {
                if (amis && !amis.has(it.imageId)) {
                    reportViolation("EC2 LaunchConfigurations should use approved AMIs.");
                }
            }),
            validateTypedResource(aws.ec2.LaunchTemplate.isInstance, (it, args, reportViolation) => {
                if (amis && it.imageId && !amis.has(it.imageId)) {
                    reportViolation("EC2 LaunchTemplates should use approved AMIs.");
                }
            }),
        ],
    };
    ```

    This is largely due to the way the `validateTypedResource` helper works to provide stronger typed resource objects to check since it only allows converting a single validation function into a strongly-typed resource. However, instead of making the property name plural (like with the previous `rules`), I've left it singular as the majority of policies will only have a single `validateResource` validation function -- but if it's simpler to implement a policy using multiple validation functions, the option is available.

    Though, we could go about strongly typing the resource in a different way which could obviate the need for multiple validation functions for such policies. More below.

 - The implementation of `ResourceValidationPolicy` has been hooked-up, but `StackValidationPolicy` has not yet. @chrsmith, want me to do it, now that your change in `pulumi/pulumi` has been merged?

 - [Here's an example](https://github.com/pulumi/examples/pull/441/files#diff-02168f5f7813feed7b3872ba18a781feR211) of a policy that is looking for potentially multiple violations, but something where you don't necessarily have to exit early when the first violation is found:

    ```typescript
        export function requireEbsEncryption(name: string, kmsKeyId?: string): ResourceValidationPolicy {
        return {
            name: name,
            description: "EBS volumes should be encrypted",
            validateResource: validateTypedResource(aws.ebs.Volume.isInstance, (it, args, reportViolation) => {
                if (!it.encrypted) {
                    reportViolation("EBS volumes should be encrypted.");
                }
                if (kmsKeyId !== undefined && it.kmsKeyId !== kmsKeyId) {
                    reportViolation(`EBS volumes should be encrypted with KMS ID '${kmsKeyId}'.`);
                }
            }),
        };
    }
    ```

## A different way to do stronger-typed resources

While the `validateTypedResource` approach to providing a strongly typed resource to program against works, it has two problems:

1. For policies that are looking for similar things over multiple resource types, it necessitates multiple validation functions (one per resource). It'd be nice if we could keep it a single validation function, and do the stronger typed filtering inside that.

2. The approach won't extend to the `validateStack` check, where you'll likely also want stronger-typed resources when looking at various resources in the stack.

One way to solve this would be to provide a helper like:

```typescript
export function asResource<TResource extends Resource>(
    isInstance: (o: any) => o is TResource,
    args: { type: string, props: Record<string, any> },
): q.ResolvedResource<TResource> | undefined {
    args.props.__pulumiType = args.type;
    return isInstance(args.props) ? args.props : undefined;
}
```

Basically, if we can convert the resource, we will return it as the stronger typed `q.ResolvedResource<TResource>`, otherwise `undefined`.

Using it in a `validateResource` function would look like:

```typescript
new PolicyPack("policy-pack-typescript", {
    policies: [{
        name: "s3-no-public-read",
        description: "Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
        validateResource: (args, reportViolation) => {
            const it = asResource(aws.s3.Bucket.isInstance, args);
            if (it && (it.acl === "public-read" || it.acl === "public-read-write")) {
                reportViolation(
                    "You cannot set public-read or public-read-write on an S3 bucket. " +
                    "Read more about ACLs here: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html");
            }
        },
    }],
});
```

And you could have multiple calls to `asResource` inside the single validation function, to cast to as many resource types as you'd like. Though, it's a bit more verbose and cumbersome than the `validateTypedResource` helper.

But the same `asResource` helper function could be used inside `validateStack` to get stronger typing for the resources there.

What do folks think?

Part of #87

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -133,9 +133,10 @@ export function validateTypedResource<TResource extends Resource>(
 ): ResourceValidation {
     return (args: ResourceValidationArgs, reportViolation: ReportViolation) => {
         args.props.__pulumiType = args.type;
-        if (typeFilter(args.props)) {
-            return validate(args.props, args, reportViolation);
+        if (typeFilter(args.props) === false) {
+            return;
         }
+        validate(args.props as q.ResolvedResource<TResource>, args, reportViolation);
     };
 }
 
@@ -150,7 +151,10 @@ export function asTypedResource<TResource extends Resource>(
     args: { type: string, props: Record<string, any> },
 ): q.ResolvedResource<TResource> | undefined {
     args.props.__pulumiType = args.type;
-    return typeFilter(args.props) ? args.props : undefined;
+    if (typeFilter(args.props) === false) {
+        return undefined;
+    }
+    return args.props as q.ResolvedResource<TResource>;
 }
 
 /**
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -121,11 +121,11 @@ export interface ResourceValidationArgs {
 
 /**
  * A helper function that returns a strongly-typed resource validation function.
- * @param filter A type guard used to determine if the args are an instance of the resource.
+ * @param typeFilter A type guard used to determine if the args are an instance of the resource.
  * @param validate A callback function that validates if the resource definition violates a policy.
  */
 export function validateTypedResource<TResource extends Resource>(
-    filter: (o: any) => o is TResource,
+    typeFilter: (o: any) => o is TResource,
     validate: (
         resource: q.ResolvedResource<TResource>,
         args: ResourceValidationArgs,
@@ -133,7 +133,7 @@ export function validateTypedResource<TResource extends Resource>(
 ): ResourceValidation {
     return (args: ResourceValidationArgs, reportViolation: ReportViolation) => {
         args.props.__pulumiType = args.type;
-        if (filter(args.props)) {
+        if (typeFilter(args.props)) {
             return validate(args.props, args, reportViolation);
         }
     };
@@ -142,15 +142,15 @@ export function validateTypedResource<TResource extends Resource>(
 /**
  * A helper function that returns `props` as a strongly-typed resolved resource based
  * on the specified `type` when `filter` returns true, otherwise `undefined` is returned.
- * @param filter A type guard used to determine if the args are an instance of the resource.
+ * @param typeFilter A type guard used to determine if the args are an instance of the resource.
  * @param args Argument bag for specifying the `type` and `props`.
  */
 export function asTypedResource<TResource extends Resource>(
-    filter: (o: any) => o is TResource,
+    typeFilter: (o: any) => o is TResource,
     args: { type: string, props: Record<string, any> },
 ): q.ResolvedResource<TResource> | undefined {
     args.props.__pulumiType = args.type;
-    return filter(args.props) ? args.props : undefined;
+    return typeFilter(args.props) ? args.props : undefined;
 }
 
 /**
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -12,17 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import * as pulumi from "@pulumi/pulumi";
 import { Resource } from "@pulumi/pulumi";
 import * as q from "@pulumi/pulumi/queryable";
 import { serve } from "./server";
 
+/**
+ * The set of arguments for constructing a PolicyPack.
+ */
 export interface PolicyPackArgs {
-    policies: Policy[];
+    /**
+     * The policies associated with a PolicyPack.
+     */
+    policies: Policies;
 }
 
+/**
+ * A PolicyPack contains one or more policies to enforce.
+ */
 export class PolicyPack {
-    private readonly policies: Policy[];
+    private readonly policies: Policies;
 
     constructor(private name: string, args: PolicyPackArgs) {
         this.policies = args.policies;
@@ -35,27 +43,6 @@ export class PolicyPack {
         serve(this.name, version, this.policies);
     }
 }
-
-/** A function that returns true if a resource definition violates some policy. */
-export type Rule = (type: string, properties: any) => void;
-
-export function typedRule<TResource extends pulumi.Resource>(
-    filter: (o: any) => o is TResource,
-    rule: (properties: q.ResolvedResource<TResource>) => void,
-): Rule {
-    return (type: string, properties: any) => {
-        properties.__pulumiType = type;
-        if (filter(properties) === false) {
-            return;
-        }
-        return rule(properties);
-    };
-}
-
-/**
- * A keyword or term to associate with a policy, such as "cost" or "security."
- */
-export type Tag = string;
 
 /**
  * Indicates the impact of a policy violation.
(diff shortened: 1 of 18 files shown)


ekrengel:
In regards to
```ts
validateResource: ResourceValidation | ResourceValidation[];
```

I think the example doesn't really make sense to me. I see these three rules as separate concerns and hence they should be separate policies.

---

I do like the "A different way to do stronger-typed resources" section. I like that it can also be used for the `validateStack`, but I am not sure how I feel about it for the `validateResource` case. I am wondering if it makes sense to have both `asResource` and `validateTypedResource`?

I think we may also need to support casting to a new interface. IE for all resources that have tags, the tags should be in this format.

justinvp:
> I see these three rules as separate concerns and hence they should be separate policies.

OK. I'll remove the array and update the examples to use separate policies.

> I am wondering if it makes sense to have both `asResource` and `validateTypedResource`?

Yeah, I think so -- they aren't mutually exclusive.

> I think we may also need to support casting to a new interface. IE for all resources that have tags, the tags should be in this format.

Hmm, yeah, we do have the following example, which kind of does this already manually without any additional built-in helpers: https://github.com/pulumi/examples/blob/567ecaf3972a48e6d5bac090c1d845d9aa1f5fbb/policy-packs/aws-advanced/compute.ts#L233-L267

clstokes:
Regarding `validateResource: ResourceValidation | ResourceValidation[];`, I think the _"Instances should use approved AMIs."_ example is a great use case for having the option to do multiple validations in a single policy. This is similar to how I've expressed checking for the existence of the same set of tags across multiple resource types too - below in the old format.

```ts
{
    name: "required-name-tag",
    description: "A 'Name' tag is required.",
    rules: [
        typedRule(aws.ec2.Instance.isInstance, it => {
            requireNameTag(it.tags);
        }),
        typedRule(aws.ec2.Vpc.isInstance, it => {
            requireNameTag(it.tags);
        }),
    ]
},
```

`asResource` seems helpful for `validateStack` but I'm failing to see how it could be used to help obviate the need for multiple validation functions. How would this work?

While we're doing API design, do we also need to consider https://github.com/pulumi/pulumi-policy/issues/78 ? As an example, how would I write a rule "instance must contain a 'Name' tag that ends with `pulumi.getStack()`"? Or will this be addressed later?

All in all, I'm good with the changes.

ekrengel:
ALSO can we come up with an example unit test? 

I think if @clstokes likes the multiple rules approach we should listen to him since he is closest to our user base.. I just want to make sure theres still an easy way to test this!

justinvp:
> `asResource` seems helpful for `validateStack` but I'm failing to see how it could be used to help obviate the need for multiple validation functions. How would this work?

It'd look like:

```typescript
{
    name: "required-name-tag",
    description: "A 'Name' tag is required.",
    validateResource: (args, reportViolation) => {
        const instance = asResource(aws.ec2.Instance.isInstance, args);
        if (instance) {
            requireNameTag(instance.tags, reportViolation);
        }
        const vpc = asResource(aws.ec2.Vpc.isInstance, args);
        if (vpc) {
            requireNameTag(vpc.tags, reportViolation);
        }
    },
},
```

Or shorter by tweaking `requireNameTag`:

```typescript
{
    name: "required-name-tag",
    description: "A 'Name' tag is required.",
    validateResource: (args, reportViolation) => {
        requireNameTag(asResource(aws.ec2.Instance.isInstance, args), reportViolation);
        requireNameTag(asResource(aws.ec2.Vpc.isInstance, args), reportViolation);
    },
},

const requireNameTag = function (resource: any | undefined, reportViolation: ReportViolation) {
    if (resource && resource.tags && resource.tags["Name"] === undefined) {
        reportViolation("A 'Name' tag is required.");
    }
};
```

justinvp:
> While we're doing API design, do we also need to consider #78 ? As an example, how would I write a rule "instance must contain a 'Name' tag that ends with `pulumi.getStack()`"? Or will this be addressed later?

This has been taken into consideration, but will be exposed later. We can add resource options, project name, and stack name to the args.

justinvp:
We discussed `validateResource: ResourceValidation | ResourceValidation[];` at the PaC Check-In meeting yesterday and decided to allow setting multiple validation functions on a single policy, as it's an easy way to check multiple types of resources in a strongly typed way using the `validateTypedResource` helper.

```typescript
{
    name: "required-name-tag",
    description: "A 'Name' tag is required.",
    validateResource: [
        validateTypedResource(aws.ec2.Instance.isInstance, (it, args, reportViolation) => {
            requireNameTag(it.tags, reportViolation);
        }),
        validateTypedResource(aws.ec2.Vpc.isInstance, (it, args, reportViolation) => {
            requireNameTag(it.tags, reportViolation);
        }),
    ],
},
```

Most policies will still have a single validation function, hence `validateResource` is singular, which should at least address the main issue with the previous API (`rules`) being plural but taking just a single rule most of the time.

sdk/nodejs/policy/policy.ts
```diff
@@ -82,10 +69,142 @@ export interface Policy {
      * proper permissions.
      */
     enforcementLevel: EnforcementLevel;
+}
+
+/**
+ * An array of Policies.
+ */
+export type Policies = (ResourceValidationPolicy | StackValidationPolicy)[];
```
ekrengel: Question bc Im still a typescript noob. Does this mean you can have an array of all `ResourceValidationPolicy` type or all `StackValidationPolicy `, but not a mix of both? I think we want to be able to define both in the same policy pack. Just want to confirm, bc this has tripped me up in the past.

sdk/nodejs/policy/policy.ts
```diff
@@ -82,10 +69,142 @@ export interface Policy {
...
+     * The type of the Resource.
+     */
+    type: string;
+
+    /**
+     * The properties of the Resource.
+     */
+    props: Record<string, any>;
+
+    // TODO: Add support for the following:
```
ekrengel: Just confirming... these are needed to validate urn/name/opts, but not needed to fulfill the current contract of PolicyEvents, correct?

sdk/nodejs/policy/policy.ts
```diff
@@ -82,10 +69,142 @@ export interface Policy {
...
+    //
+    // urn: string;
+    // name: string;
+    // opts: PolicyResourceOptions;
+}
+
+/**
+ * A helper function that returns a strongly-typed resource validation function.
+ * @param filter A type guard used to determine if the args are an instance of the resource.
```
ekrengel: Would `typeFilter` or `typeGuard` make more sense here?

sdk/nodejs/policy/policy.ts
```diff
@@ -82,10 +69,142 @@ export interface Policy {
...
+        if (filter(args.props)) {
+            return validate(args.props, args, reportViolation);
+        }
+    };
+}
 
+/**
+ * A helper function that returns `props` as a strongly-typed resolved resource based
+ * on the specified `type` when `filter` returns true, otherwise `undefined` is returned.
+ * @param filter A type guard used to determine if the args are an instance of the resource.
```
ekrengel: Same Q here in regards to naming.

sdk/nodejs/policy/server.ts
```diff
@@ -90,7 +96,7 @@ async function getPluginInfoRpc(call: any, callback: any): Promise<void> {
 }
 
 // analyze is the RPC call that will analyze an individual resource, one at a time (i.e., check).
-function makeAnalyzeRpcFun(policyPackName: string, policyPackVersion: string, policies: Policy[]) {
+function makeAnalyzeRpcFun(policyPackName: string, policyPackVersion: string, policies: Policies) {
```
ekrengel: How do we test if this is working as expected?

sdk/nodejs/policy/policy.ts
justinvp: Let's go with `typeFilter` which is also the naming Alex used [here](https://github.com/pulumi/pulumi/blob/23a3acc46f0df6deec4443d0f107400311c2f960/sdk/nodejs/runtime/resource.ts#L563).

sdk/nodejs/policy/policy.ts
justinvp: Correct.

sdk/nodejs/policy/policy.ts
justinvp: Mix of both. I created a [Playground](http://www.typescriptlang.org/play/?ssl=47&ssc=1&pln=7&pc=1#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgBQgBsFsBPfHAawEEoBzAZzgG8AoOTuSE7BYRgC4CxUv0YBuNgF82bUJFhxsRTI2aFeFaqw5cwUBADdMqOFGCYAJhCREy3UXwHDNYgVL2dstxjCgArtgw0AAUBsameEiYALbAwn6GSPQANHCYDEIiWpTYtFkAlLpcpXAwABYIjAB0PO7MALwZWXVO4lKlsrLy4NDwMGRgeG7OTXChAEoCEAFQuABqmCRWpgi2ow4APnAAyjDUSytrG05khQDaALqeCv3lQ3gAokjo0LjxKAAywEbARHBmgAiaxGarQMhAuA7IGxTBIVYhKCQ259JTIVAYHAjM4lLgxeKJfzIeidLhWATYQxgGDrJBE5KkrxwYCvd7AT4wH5-IjCF5veYc1lc37-KQ9O7olBoLC4ODTRizQVHBCIumbFkgVAIjS49ilEzHVAKpW4YQAIwgxEsSAkcAA9Pa4ALEEQiAEkms-tw5pBGAIZHJJfAMTLsXsDvkVWrTlpNdqrLq4-quIbVVF9tQLVaiDa7Y7ndBXe7PbTvWBfRB-YxA2wfEg-I5eOJAXAkMAAO45UjafKhIFgMBA9Ipzj1MbCC7M0qjsqlAkJOBAiyKua4AC0aZjSHX48hqWnc7gFMYVIQNLpwiBJ+ww8Pc9ZAo+wu5-yvoPByLvR7KW6iJrXRd-ACYB7zgaQDx-WcfwXK8-GoTdlnTWlbF3M5vx-clKWpFD6SXG8MMwlk2UFTlX15JcP0VL9IKIuA-1QTN8mEYDQJ-CDmSuA9pEKKQgA) that demonstrates.

If it was all `ResourceValidationPolicy` or all `StackValidationPolicy` (which we don't want), it'd be typed as `ResourceValidationPolicy[] | StackValidationPolicy[]`.

sdk/nodejs/policy/server.ts
justinvp: I've manually tested it, but I will add some actual integration tests.

justinvp:
@ekrengel, I added some basic integration tests. Definitely room to add a lot more tests here, like validating unknowns, adding a Python program, etc., but we can add those subsequently as part of #9.

ekrengel:
>  I added some basic integration tests. 

Awesome thanks for taking the time to do this :)

PR #135, opened 2019-11-05
Hook up `StackValidationPolicy`

### LINKED ISSUE #105: Policy SDK to allow you to aggregate resources
Per Cameron:

Can I aggregate data across rule executions?

And really what I'm getting at is, can I do a cost estimation of a pulumi stack? Effectively, count the t3.mediums in a stack, multiply by $0.0X, and then print a warning if the sum is greater than $XX?

justinvp:
This hooks up the changes in https://github.com/pulumi/pulumi/pull/3366 to the new `validateStack` API.

I ran into two issues while working on this that we should discuss before merging this:

1. During previews, the outputs are always empty. It took me a while to figure this out. I thought I'd screwed up the gRPC calls, but as soon as I added `--skip-preview` to the `pulumi up` commands in the integration tests I started seeing the outputs.

    I think this is going to be confusing for folks writing policies. At the very least, should we consider providing an `isPreview` property on the args, that could be checked to see if the policy is being run during a preview vs. update? Or only run stack validation policies during updates (I don't think we want to do that, since there is a class of policies you might still want to fail during previews, such as basic cost analysis based on the type, or only allowing a certain number of resources)?

  2. We're [currently _only_ passing the resources outputs](https://github.com/pulumi/pulumi/blob/a9db935ccc96b8d35c172c7ca31ebc4387738d0a/pkg/resource/deploy/plan_executor.go#L277-L279). We really should be sending both the outputs _and_ inputs (merging the two like we do [here](https://github.com/pulumi/pulumi/blob/29049ecaf31ebe9f66870de415e7cda636ebc325/sdk/nodejs/runtime/resource.ts#L470-L475)). Either that or have separate `inputs` and `outputs` properties rather than a single `props`.

cc @chrsmith, @ekrengel

Fixes #105

### DIFF
--- a/sdk/nodejs/policy/package.json
+++ b/sdk/nodejs/policy/package.json
@@ -7,7 +7,7 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-policy",
     "dependencies": {
-        "@pulumi/pulumi": "0.17.23",
+        "@pulumi/pulumi": "1.4.1",
         "google-protobuf": "^3.5.0",
         "grpc": "^1.20.2",
         "protobufjs": "^6.8.6"
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -23,9 +23,12 @@ import { deserializeProperties } from "./deserialize";
 import {
     Policies,
     Policy,
+    PolicyResource,
     ReportViolation,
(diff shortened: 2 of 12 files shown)


justinvp:
(Marking as no-changelog-required, as the changelog already mentions the new `StackValidationPolicy` API)

ekrengel:
Did you want someone to review this?

PR #139, opened 2019-11-11
Don't used `Resolved` for strongly-typed resource policies

### LINKED ISSUE #116: Don't used `Resolved` for policy types
Policy rules act on "fully resolved" resource inputs—in other words, the resource definition, after all `Output<T>` has been fully resolved to the underlying values.

To create this "fully resolved input shape", we currently expose `pulumi.queryable.ResolvedResource<T>`. This takes some `T extends pulumi.Resource` and -> the "resolve" shape, which—we thought—we could use in policy rules.

This is actually not quite right. For example, the `Resource` class is the fully-materialized version of the _output_ shape of the resource. But we want the _input_ shape. Probably this should be derived from the type of the args passed to the `Resource` constructor when it's created. This was originally used for `pulumi query`, which does use the output shape.

Quoting from [here](https://github.com/pulumi/pulumi-policy/issues/92) for more context:

> We actually need to transition away from `Resolved` and towards something that unwraps the `Input` shapes—`Resolved` on the `CustomResource` types (among other things) does not fully capture nullable fields, and includes output-only fields that will never be populated in an update. We meant to do this but haven't gotten to it yet because (1) the unknowns issue was much more urgent, and (2) we had to switch focus to v1 work. But, this is certainly on the list.
> 
> I thought we had an issue for this, but—somehow there isn't? I'll file one.

justinvp:
We currently use `pulumi.queryable.ResolvedResource<T>` in the strongly-typed `validateTypedResource` helper, but this represents the "output" shape of the object when it really should be the "input" shape, since these policies are receiving the resource inputs.

This PR modifies `validateTypedResource` so that it provides the "input" shape of the resource, i.e. the args type used when creating the resource.

Instead of passing a `typeFilter` type guard, the class itself is passed. From this, we can infer the resource's args type, based on its constructor parameters, which represents the "input" shape we want. We use [`Unwrap<TArgs>`](https://github.com/pulumi/pulumi/blob/1a6897ba262c0668340fa5e62a26f89cc538ec41/sdk/nodejs/output.ts#L460-L490) so that the properties are typed as `T` instead of `Input<T>`.

Before:

```typescript
// Passing the `isInstance` function.
validateResource: validateTypedResource(aws.s3.Bucket.isInstance, (bucket, args, reportViolation) => {
    // Here `bucket` is the "resolved" output shape, e.g. `aws.s3.Bucket` with all properties typed as `T` instead of `Output<T>`.
});
```

After:

```typescript
// Passing the class itself.
validateResource: validateTypedResource(aws.s3.Bucket, (bucket, args, reportViolation) => {
    // Here `bucket` is the "unwrapped" input shape, e.g. `aws.s3.BucketArgs` with all properties typed as `T` instead of `Input<T>`.
}):
```

Updated examples here: https://github.com/pulumi/examples/pull/455

Fixes #116

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Resource } from "@pulumi/pulumi";
+import { Resource, Unwrap } from "@pulumi/pulumi";
 import * as q from "@pulumi/pulumi/queryable";
 import { serve } from "./server";
 
@@ -121,22 +121,26 @@ export interface ResourceValidationArgs {
 
 /**
  * A helper function that returns a strongly-typed resource validation function.
- * @param typeFilter A type guard used to determine if the args are an instance of the resource.
+ * @param resourceClass A resource class used to filter this check to only resources of the specified class and
+ * determine the appropriate strongly-typed `TArg` type to use for the resource.
  * @param validate A callback function that validates if the resource definition violates a policy.
  */
-export function validateTypedResource<TResource extends Resource>(
-    typeFilter: (o: any) => o is TResource,
+export function validateTypedResource<TResource extends Resource, TArgs>(
+    resourceClass: { new(name: string, args: TArgs, ...rest: any[]): TResource },
     validate: (
-        resource: q.ResolvedResource<TResource>,
+        props: NonNullable<Unwrap<TArgs>>,
         args: ResourceValidationArgs,
         reportViolation: ReportViolation) => Promise<void> | void,
 ): ResourceValidation {
     return (args: ResourceValidationArgs, reportViolation: ReportViolation) => {
-        args.props.__pulumiType = args.type;
-        if (typeFilter(args.props) === false) {
+        const isInstance = (<any>resourceClass).isInstance;
+        if (!isInstance || typeof isInstance !== "function") {
             return;
         }
-        validate(args.props as q.ResolvedResource<TResource>, args, reportViolation);
+        if (isInstance({ __pulumiType: args.type }) !== true) {
+            return;
+        }
+        validate(args.props as NonNullable<Unwrap<TArgs>>, args, reportViolation);
     };
 }
 
@@ -150,8 +154,7 @@ export function asTypedResource<TResource extends Resource>(
     typeFilter: (o: any) => o is TResource,
     args: { type: string, props: Record<string, any> },
 ): q.ResolvedResource<TResource> | undefined {
-    args.props.__pulumiType = args.type;
-    if (typeFilter(args.props) === false) {
+    if (typeFilter({ __pulumiType: args.type }) !== true) {
         return undefined;
     }
     return args.props as q.ResolvedResource<TResource>;
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -145,22 +145,6 @@ export function validateTypedResource<TResource extends Resource, TArgs>(
 }
 
 /**
- * A helper function that returns `props` as a strongly-typed resolved resource based
- * on the specified `type` when `filter` returns true, otherwise `undefined` is returned.
- * @param typeFilter A type guard used to determine if the args are an instance of the resource.
- * @param args Argument bag for specifying the `type` and `props`.
- */
-export function asTypedResource<TResource extends Resource>(
-    typeFilter: (o: any) => o is TResource,
-    args: { type: string, props: Record<string, any> },
-): q.ResolvedResource<TResource> | undefined {
-    if (typeFilter({ __pulumiType: args.type }) !== true) {
-        return undefined;
-    }
-    return args.props as q.ResolvedResource<TResource>;
-}
-
-/**
  * StackValidationPolicy is a policy that validates a stack.
  */
 export interface StackValidationPolicy extends Policy {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@
           name: "s3-no-public-read",
           description: "Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
           enforcementLevel: "mandatory",
-          validateResource: validateTypedResource(aws.s3.Bucket.isInstance, (it, args, reportViolation) => {
-              if (it.acl === "public-read" || it.acl === "public-read-write") {
+          validateResource: validateTypedResource(aws.s3.Bucket, (bucket, args, reportViolation) => {
+              if (bucket.acl === "public-read" || bucket.acl === "public-read-write") {
                   reportViolation(
                       "You cannot set public-read or public-read-write on an S3 bucket. " +
                       "Read more about ACLs here: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html");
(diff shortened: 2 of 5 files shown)


sdk/nodejs/policy/policy.ts
```diff
@@ -121,42 +121,29 @@ export interface ResourceValidationArgs {
...
     };
 }
 
-/**
- * A helper function that returns `props` as a strongly-typed resolved resource based
- * on the specified `type` when `filter` returns true, otherwise `undefined` is returned.
- * @param typeFilter A type guard used to determine if the args are an instance of the resource.
- * @param args Argument bag for specifying the `type` and `props`.
- */
-export function asTypedResource<TResource extends Resource>(
```
justinvp: This helper is no longer generally useful. For `validateResource` we want the "input" shape, but for `validateStack`, we want the "output" shape.

It'd be best to simplify things here by removing this. For `validateResource`, if you want to write your validation function in a strongly-typed way, use `validateTypedResource`. For `validateStack`, if you want to filter/cast resources, use `@pulumi/query`. As a follow-up, I'll make sure we have an example of how to use `@pulumi/query` with `validateStack`.

sdk/nodejs/policy/policy.ts
```diff
@@ -121,42 +121,29 @@ export interface ResourceValidationArgs {
...
  * A helper function that returns a strongly-typed resource validation function.
- * @param typeFilter A type guard used to determine if the args are an instance of the resource.
+ * @param resourceClass A resource class used to filter this check to only resources of the specified class and
+ * determine the appropriate strongly-typed `TArg` type to use for the resource.
  * @param validate A callback function that validates if the resource definition violates a policy.
  */
-export function validateTypedResource<TResource extends Resource>(
-    typeFilter: (o: any) => o is TResource,
+export function validateTypedResource<TResource extends Resource, TArgs>(
+    resourceClass: { new(name: string, args: TArgs, ...rest: any[]): TResource },
```
justinvp: When you pass `aws.s3.Bucket` here, the type of its args (`TArgs`) is inferred from the constructor parameters. This relies on the pattern we generally follow for our resource types, where the first parameter is the name of the resource and the second parameter is it's args bag.

sdk/nodejs/policy/policy.ts
```diff
@@ -121,42 +121,29 @@ export interface ResourceValidationArgs {
...
     validate: (
-        resource: q.ResolvedResource<TResource>,
+        props: NonNullable<Unwrap<TArgs>>,
```
justinvp: The resource's `TArgs` constructor parameter is typically optional, so use TypeScript's `NonNullable<T>` it isn't `TArgs | undefined` here. We also use [`Unwrap<T>`](https://github.com/pulumi/pulumi/blob/1a6897ba262c0668340fa5e62a26f89cc538ec41/sdk/nodejs/output.ts#L460-L490) so that the properties are typed as the underlying type instead of `Input<T>` (e.g. `string` instead of `Input<string>`.

sdk/nodejs/policy/policy.ts
ekrengel: So if we eve deviate from this pattern we'd be in trouble here? I wondering if theres anywhere we do deviate / will this also work in other languages?

sdk/nodejs/policy/policy.ts
ekrengel: Definitely interested in seeing what this looks like!

sdk/nodejs/policy/policy.ts
justinvp: > So if we ever deviate from this pattern we'd be in trouble here?

If we deviate from the pattern, the resource won't be able to make use of the strongly-typed helper. You'd have to use a regular untyped validation function. It does make me a little uneasy, but this should work and provide the nicer experience we want for the vast majority of the resources we provide (I suspect ~99%), and for the cases where it can't be used (can we can't just go fix the resource to follow the pattern), you can still easily workaround using a regular untyped validation function.

> I wondering if theres anywhere we do deviate / will this also work in other languages?

I will look. I suspect that if there are any places where we deviate, it would be in our high-level libraries like AWSX that have resources that inherit from `ComponentResource` rather than `CustomResource`. There's a separate issue (#140) with `ComponentResource` in that it doesn't use the `__pulumiType` pattern we're using, but I will fix that subsequently.

justinvp: As for other languages, I still want to do some prototyping, but I don't think we'll be able to automatically infer the args type as nicely as we can with TypeScript.

justinvp: I took a look at resources in `awsx` and `eks`... There are a handful of resources that extend `ComponentResource` that have extra parameters that don't work with the pattern:

```typescript
export class StepScalingPolicy extends pulumi.ComponentResource {
    constructor(name: string, group: AutoScalingGroup,
                args: StepScalingPolicyArgs, opts: pulumi.ComponentResourceOptions = {}) {}
}

export class Subnet extends pulumi.ComponentResource {
    constructor(name: string, vpc: x.ec2.Vpc, args: SubnetArgs, opts?: pulumi.ComponentResourceOptions);
    constructor(name: string, vpc: x.ec2.Vpc, args: ExistingSubnetArgs, opts?: pulumi.ComponentResourceOptions);
    constructor(name: string, vpc: x.ec2.Vpc, args: SubnetArgs | ExistingSubnetArgs, opts: pulumi.ComponentResourceOptions = {}) {}
}

export class EgressSecurityGroupRule extends SecurityGroupRule {
    constructor(name: string, securityGroup: x.ec2.SecurityGroup,
                args: SimpleSecurityGroupRuleArgs | EgressSecurityGroupRuleArgs,
                opts?: pulumi.ComponentResourceOptions) {}
}

export class IngressSecurityGroupRule extends SecurityGroupRule {
    constructor(name: string, securityGroup: x.ec2.SecurityGroup,
                args: SimpleSecurityGroupRuleArgs | IngressSecurityGroupRuleArgs,
                opts?: pulumi.ComponentResourceOptions) {}
}

export class LifecyclePolicy extends aws.ecr.LifecyclePolicy {
    constructor(name: string, repository: aws.ecr.Repository, args?: LifecyclePolicyArgs, opts: pulumi.ComponentResourceOptions = {}) {}
}

export class ListenerRule extends pulumi.ComponentResource {
    constructor(name: string, listener: x.lb.Listener,
                args: ListenerRuleArgs, opts: pulumi.ComponentResourceOptions = {}) {
}

export class TargetGroupAttachment extends pulumi.ComponentResource {
    constructor(name: string, targetGroup: mod.TargetGroup, args: mod.LoadBalancerTarget, opts: pulumi.ComponentResourceOptions = {}) {
}
```

This is a pretty small number compared to the vast number of resources from TF and Kubernetes that follow the pattern. There may be some room for another strongly-typed helper that'd fit in-between the low-level untyped validation method and the constructor pattern-based `validateTypedResource` helper, that could be used for such types. Will think about this as part of addressing #140.

In eks, there was a dynamic resource, but we already couldn't filter those because [they all have the same type](https://github.com/pulumi/pulumi/blob/137fd54f1cdd5ff901ddba0f281039bf3b9a5e01/sdk/nodejs/dynamic/index.ts#L189) (`"pulumi-nodejs:dynamic:Resource"`). We should think about whether we want to make it easy to write policies for dynamic resources in a strongly-typed way separately.

```
export class VpcCni extends pulumi.dynamic.Resource {
    constructor(name: string, kubeconfig: pulumi.Input<any>, args?: VpcCniOptions, opts?: pulumi.CustomResourceOptions) {}
}
```

ekrengel: Is this something we should call out in docs as a limitation? 🤔

PR #143, opened 2019-11-13
Add way to do strong typing for atypical resources

justinvp:
The vast majority of resources follow the same order of constructor parameters, which allows us to infer the args type (input shape) used when creating the resource. However, there are resources that don't follow the pattern of constructor parameters. This change provides a way to still have strongly-typed validation checks for such resources, you just have to specify the types explicitly.

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import { Resource, Unwrap } from "@pulumi/pulumi";
-import * as q from "@pulumi/pulumi/queryable";
 import { serve } from "./server";
 
 /**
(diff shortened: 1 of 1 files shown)


ekrengel (commented):
A bit confused. I think this means if we dont follow the typical pattern we can use 
```validateTypedResource<awsx.ec2.Subnet, awsx.ec2.SubnetArgs>(awsx.ec2.Subnet, (s, args, report) => { ... })```
instead?

Its not super clear to me when to use one verse the other / how would others know when they need to use one vs the other?

--- 
Also lets say I make an API gateway with AWSX and it violates a policy for an AWS gateway. Would the resource fail the policy?

justinvp:
> A bit confused. I think this means if we dont follow the typical pattern we can use
> `validateTypedResource<awsx.ec2.Subnet, awsx.ec2.SubnetArgs>(awsx.ec2.Subnet, (s, args, report) => { ... })`
> instead?

Right. This is really just providing an escape hatch to make it possible to use `validateTypedResource` with resources that don't have the typical constructor parameters in the typical order. In such cases, the compiler won't be able to infer the resource args type. This change makes it possible to specify those types explicitly.

> Its not super clear to me when to use one verse the other / how would others know when they need to use one vs the other?

The vast majority of the time you won't have to know or care about doing this. For almost all the resources we provide, the compiler will happily infer the types automatically, so you don't have to specify them explicitly (i.e. you can write `validateTypedResource(aws.s3.Bucket, ...)` instead of having to specify them like `validateTypedResource<aws.s3.Bucket, aws.s3.BucketArgs>(aws.s3.Bucket, ...)`). But for the cases where it can't infer the args type, this provides a way for the types to be specified explicitly. We should definitely document it.

> Also lets say I make an API gateway with AWSX and it violates a policy for an AWS gateway. Would the resource fail the policy?

What do you mean?

justinvp:
I'm going to think about this a little more. It is a way we could allow atypical resources to work with the strongly-typed helper. You'd rarely encounter resources where you'd need to do it, but having to specify the resource type twice isn't ideal. In the meantime, I'll close this PR.

PR #144, opened 2019-11-13
Update CHANGELOG and README for v0.2.0

justinvp:


README.md
```diff
@@ -84,68 +85,66 @@ You can find more example Policy Packs in the [examples repo](https://github.com
...
-```sh
-$ PULUMI_DEBUG_COMMANDS=true pulumi preview \
---policy-pack ~/policy-pack-typescript
-Previewing update (dev):
+1. We then run the `pulumi preview` command again and this time get an error message indicating we failed the preview because of a policy violation.
 
-     Type                 Name          Plan       Info
- +   pulumi:pulumi:Stack  test-dev  	create     1 error
- +   └─ aws:s3:Bucket     my-bucket     create     1 error
+    ```sh
```
ekrengel: Do the numbers still get generate correctly? I found when I moved these over before the number would restart.

justinvp: Good catch. Without indenting, the numbers always stay "1". Fixed.

PR #148, opened 2019-11-16
Cleanup error message when running two gRPCs in same proc

### LINKED ISSUE #106: Unhelpful error when publishing with two PolicyPack objects
If I try to publish a project with more than one `PolicyPack` object, I get a confusing error, but the publish seems to complete successfully. 

As an aside, should I be able to publish with multiple packs like this? If not, we should document this and provide a helpful error message.

#### Code to Reproduce
```typescript
import * as aws from "@pulumi/aws";
import { PolicyPack, typedRule } from "@pulumi/policy";
import * as assert from "assert";

const cost = new PolicyPack("cost", {
    policies: [
        {
            name: "prohibited-elasticbeanstalk",
            description: "Use of Elastic Beanstalk is prohibited.",
            rules: (type: string) => {
                assert(type.startsWith("aws:elasticbeanstalk") === false);
            },
        },
    ],
});

const security = new PolicyPack("security", {
    policies: [
        {
            name: "discouraged-ec2-public-ip-address2",
            description: "Associating public IP addresses is discouraged.",
            enforcementLevel: "advisory",
            rules: typedRule(aws.ec2.Instance.isInstance, it => {
                assert(it.associatePublicIpAddress === false);
            }),
        },
    ]
});
```

#### Output
```
15:56:33 [policies] $ pulumi policy publish clstokes/ec2-4
Obtaining policy metadata from policy plugin
error: [runtime] Running program '/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies' failed with an unhandled exception:

    Error: Only one policy gRPC server can run per process

    at Object.serve (/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies/node_modules/@pulumi/policy/server.js:52:15)

    at new PolicyPack (/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies/node_modules/@pulumi/policy/policy.js:25:18)

    at Object.<anonymous> (/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies/index.ts:18:18)

    at Module._compile (internal/modules/cjs/loader.js:936:30)

    at Module.m._compile (/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies/node_modules/ts-node/src/index.ts:439:23)

    at Module._extensions..js (internal/modules/cjs/loader.js:947:10)

    at Object.require.extensions.<computed> [as .ts] (/Users/clstokes/cc/pulumi/sales/demos/aws-ts-webserver/policies/node_modules/ts-node/src/index.ts:442:12)

    at Module.load (internal/modules/cjs/loader.js:790:32)

    at Function.Module._load (internal/modules/cjs/loader.js:703:12)

    at Module.require (internal/modules/cjs/loader.js:830:19)

Compressing policy pack
Uploading policy pack to Pulumi service
Publishing as ec2-4
Published as version 1
15:57:26 [policies] $
```

chrsmith:
Because of the specific way that the Pulumi CLI loads plugins, and the way the PAC SDK sets up its gRPC server, we don't allow you to define multiple policy packs within the same process.

While it's conceivable that we would want to support this in the future -- e.g. have a single "policy program" that defined multiple "policy packs" -- this limit seems like a reasonable thing to enforce for the time being.

This PR simply changes the way we report this error so that it is more user friendly when the user sees it.

Rather than throwing an exception with a developer-facing error message, we instead write to STDERR and kill the process. This prevents a large/confusing callstack from being rendered, as well as makes the problem easier to understand.

This is the output you will see from the Pulumi CLI. Since the process gets killed, we fail to complete standing up/registering the gRPC handshake and/or will have some other failure communicating with the plugin. But the policy process's STDOUT will get printed to the Pulumi program's output.

```
$ pulumi preview \
    --policy-pack=${HOME}/go/src/github.com/pulumi/pulumi-awsguard/
Previewing update (pac-testing):

     Type                 Name                   Plan     Info
     pulumi:pulumi:Stack  resources-pac-testing           1 error; 1 message
 
Diagnostics:
  pulumi:pulumi:Stack (resources-pac-testing):
    error: policy pack "/Users/chris/go/src/github.com/pulumi/pulumi-awsguard" failed to start: /Users/chris/go/src/github.com/pulumi/pulumi-awsguard (analyzer) plugin [/Users/chris/.pulumi/bin/pulumi-analyzer-policy] did not begin responding to RPC connections
 
    Already serving policy pack 'pulumi-awsguard'. Only one policy pack may be defined per-process.
 
Permalink: https://app.pulumi.com/chrsmith/resources/pac-testing/previews/2453e2a3-9d50-4ea3-bd06-89194fd72ad2
```

The error could still be improved by touching up the code in the `pulumi/pulumi` repo around how we load plug-ins and surface their output. But as a first step I think this is enough to fix #106

### DIFF
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -14,10 +14,7 @@
 
 const grpc = require("grpc");
 const analyzerrpc = require("@pulumi/pulumi/proto/analyzer_grpc_pb.js");
-const structproto = require("google-protobuf/google/protobuf/struct_pb.js");
 const plugproto = require("@pulumi/pulumi/proto/plugin_pb.js");
-
-import { AssertionError } from "assert";
 
 import { deserializeProperties } from "./deserialize";
 import {
@@ -35,10 +32,11 @@ import {
     Diagnostic,
     makeAnalyzeResponse,
     makeAnalyzerInfo,
-    mapEnforcementLevel,
 } from "./protoutil";
 import { unknownCheckingProxy, UnknownValueError } from "./proxy";
 import { version } from "./version";
+
+import * as process from "process";
 
 // ------------------------------------------------------------------------------------------------
 
@@ -48,14 +46,18 @@ import { version } from "./version";
 
 // ------------------------------------------------------------------------------------------------
 
-let serving = false;
+// Name of the policy pack currently being served, if applicable.
+let servingPolicyPack: string | undefined = undefined;
 
 export function serve(policyPackName: string, policyPackVersion: string, policies: Policies): void {
-    if (serving !== false) {
-        throw Error("Only one policy gRPC server can run per process");
+    if (servingPolicyPack) {
+        // We only support running one gRPC instance at a time. (Since the Pulumi CLI is looking for a single
+        // PID to be written to STDOUT.) So we just print an error and kill the process if a second policy pack
+        // is about to be served.
+        console.error(`Already serving policy pack '${servingPolicyPack}'. Only one policy pack may be defined per-process.`);
+        process.exit(1);
     }
-
-    serving = true;
+    servingPolicyPack = policyPackName;
 
     // Finally connect up the gRPC client/server and listen for incoming requests.
     const server = new grpc.Server();


sdk/nodejs/policy/server.ts
```diff
@@ -48,14 +46,18 @@ import { version } from "./version";
...
 // ------------------------------------------------------------------------------------------------
 
-let serving = false;
+// Name of the policy pack currently being served, if applicable.
+let servingPolicyPack: string | undefined = undefined;
 
 export function serve(policyPackName: string, policyPackVersion: string, policies: Policies): void {
-    if (serving !== false) {
-        throw Error("Only one policy gRPC server can run per process");
+    if (servingPolicyPack) {
```
justinvp: We don't currently do any validation of `policyPackName` in this library. We probably should just do the same validation that the service does inside the `PolicyPack` constructor. Otherwise, right now, I could pass `undefined`, `null`, or `""`, which would throw off this logic.

chrsmith: SGTM, I'll send that out in a separate PR.

PR #150, opened 2019-11-19
Cleanup violation messages

chrsmith:
Make a small change to format of violation errors as well as add an error if the policy pack name is invalid.

This allows the user to call `reportViolation("")` or `reportViolation(undefined)` without causing weird output down the line.

... what do you think about this? It's also possible the policy's description couldn't be provided too. Should we add some hook to validate that on startup, every provided policy is "well-formed" with a name, description, and valid enforcement level?

We'd need to put the same validation in the Pulumi CLI and service. So perhaps we shouldn't bother trying to enforce these types of things in the SDK layer?

### DIFF
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -49,6 +49,9 @@ import * as process from "process";
 // Name of the policy pack currently being served, if applicable.
 let servingPolicyPack: string | undefined = undefined;
 
+// Regular expression for the policy pack name.
+const packNameRE = "^[a-zA-Z0-9-_.]{1,100}$";
+
 /**
   * Starts the gRPC server to communicate with the Pulumi CLI client for analyzing resources.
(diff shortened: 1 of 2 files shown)


chrsmith:
@justinvp , @ekrengel going ahead and merging this. I am not sure if we want to put this in the `CHANGELOG.md`, but we may need to update tests in `@pulumi/awsguard` when we update to the newer version of this library. (Since we change the format of the diagnostic messages we emit from the Policy Pack server.)

We no longer print the name of the policy/rule in the diagnostic. Just the description and message.

```diff
pulumi-nodejs:dynamic:Resource (c):
-  mandatory: [dynamic-no-state-with-value-3] Prohibits setting state to 3 on dynamic resources.
+ mandatory: Prohibits setting state to 3 on dynamic resources.
'state' must not have the value 3.
```

PR #152, opened 2019-11-22
Add coverage for passing relative paths to --policy-pack

justinvp:
Depends on https://github.com/pulumi/pulumi/pull/3565 (will fail until that's merged and released, or we pin to a dev build of the CLI).


ekrengel (approved):
I think it may be cleaner if we use a table test here.

justinvp:
Closing stale PR

PR #156, opened 2019-11-25
Add "disabled" value to EnforcementLevel

justinvp:
This allows disabling policies. A disabled policy is treated as if it wasn't defined, which means it will be excluded from analyzer info and when analyzing resources and the stack (i.e. the CLI and service don't need to know about `"disabled"`).

### DIFF
--- a/sdk/nodejs/policy/protoutil.ts
+++ b/sdk/nodejs/policy/protoutil.ts
@@ -130,7 +130,17 @@ export function mapEnforcementLevel(el: EnforcementLevel) {
             return analyzerproto.EnforcementLevel.ADVISORY;
         case "mandatory":
             return analyzerproto.EnforcementLevel.MANDATORY;
+        // Disabled is treated as if the policy was not defined, so the value should not escape over GRPC.
+        case "disabled":
+            throw new Error("'disabled' should not escape the GRPC boundary");
         default:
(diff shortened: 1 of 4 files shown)


sdk/nodejs/policy/protoutil.ts
```diff
@@ -130,7 +130,17 @@ export function mapEnforcementLevel(el: EnforcementLevel) {
             return analyzerproto.EnforcementLevel.ADVISORY;
         case "mandatory":
             return analyzerproto.EnforcementLevel.MANDATORY;
+        // Disabled is treated as if the policy was not defined, so the value should not escape over GRPC.
+        case "disabled":
+            throw new Error("'disabled' should not escape the GRPC boundary");
```
chrsmith: As a first step this is fine, but I think we should wire this through the CLI layer and add some comments in the protobuf. Since otherwise an alternate implementation could "leak" the "disabled" bit. Similarly, when we do hook this option up to the Pulumi Service UI, we will need to pass the "disabled" state through the CLI to the policy pack.

Read: it's fine for now, but probably not sufficient for the long run.

ekrengel: Yeah we talked about this but for now its nice to not to have to touch the service!

PR #169, opened 2019-12-05
Test Config/getProject/getStack/isDryRun in policy packs

justinvp:
Depends on https://github.com/pulumi/pulumi/pull/3612. The tests will fail until we have a new release of the CLI and `@pulumi/pulumi`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## HEAD (Unreleased)
 
+### Improvements
+
+- Add support for using `Config`, `getProject()`, `getStack()`, and `isDryRun()` from Policy Packs
+  via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 of the Pulumi SDK).
+
 ## 0.3.0 (2019-11-26)
 
 ### Improvements
--- a/sdk/nodejs/policy/package.json
+++ b/sdk/nodejs/policy/package.json
@@ -7,7 +7,7 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-policy",
     "dependencies": {
-        "@pulumi/pulumi": "^1.6.1",
+        "@pulumi/pulumi": "^1.8.0",
         "google-protobuf": "^3.5.0",
         "grpc": "^1.20.2",
         "protobufjs": "^6.8.6"
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -60,7 +60,11 @@ func runPolicyPackIntegrationTest(
 	}
 	rootDir := path.Join(cwd, testDirName)
 
+	// The Pulumi project name matches the test dir name in these tests.
+	os.Setenv("PULUMI_TEST_PROJECT", testDirName)
+
 	stackName := fmt.Sprintf("%s-%d", testDirName, time.Now().Unix()%100000)
+	os.Setenv("PULUMI_TEST_STACK", stackName)
 
 	// Copy the root directory to /tmp and run various operations within that directory.
 	e := ptesting.NewEnvironment(t)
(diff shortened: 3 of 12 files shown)


tests/integration/basic_data/policy-pack/PulumiPolicy.yaml
```diff
@@ -0,0 +1,2 @@
+description: A Policy Pack to test certain data is available to policy packs
```
chrsmith: Nit: Should this end with a period? IMHO, it looks like it should.

tests/integration/basic_data/policy-pack/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import * as aws from "@pulumi/aws";
+import { PolicyPack } from "@pulumi/policy";
+
+new PolicyPack("basic-data-policy", {
+    policies: [
+        {
+            name: "basic-data-resource-validation",
+            description: "Verifies data during resource validation.",
+            enforcementLevel: "mandatory",
+            validateResource: (_, __) => {
```
chrsmith: Nit: IMHO `a, r` or `x, y` would be preferable over `_, __`.

tests/integration/basic_data/policy-pack/PulumiPolicy.yaml
chrsmith: "certain data" isn't especially useful. (Similarly `basic_data` is equally vague.)

What do you think about calling the test "runtime_data" or "pulumi_sdk"?

tests/integration/basic_data/policy-pack/index.ts
```diff
@@ -0,0 +1,52 @@
...
+            validateStack: (_, __) => {
+                verifyData();
+            },
+        },
+    ],
+});
+
+function verifyData() {
+    // Verify isDryRun().
+    assert.strictEqual(pulumi.runtime.isDryRun(), true, "'isDryRun()' not the expected value");
```
chrsmith: Can we be more precise here?

It isn't that `isDryRun()` is _always_ expected to be set. It's that our test harness only performs previews. I think that distinction is important enough to call out with a comment, or maybe even having wiring through another test scenario where we verify these fields during a proper `pulumi up`.

tests/integration/basic_data/program/index.ts
```diff
@@ -0,0 +1 @@
+// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
```
chrsmith: Nit: Maybe add a comment that the Pulumi program doesn't _do_ anything, and that the tests just verify the properties of _all_ resources created. (Which in this case would just be the stack resource.)

Or perhaps just `console.log("[this Pulumi program intentionally left blank]");`?

tests/integration/basic_data/program/resource.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
+
+import * as pulumi from "@pulumi/pulumi";
```
chrsmith: It doesn't look like this is called from `program/index.ts`. Is this still necessary?

tests/integration/basic_data/policy-pack/PulumiPolicy.yaml
ekrengel: I dont think we typically have this end in a period, but agree that follows suit to our other comments/text throughout our repos. 

Should it be `available to Policy Packs`?

tests/integration/basic_data/policy-pack/index.ts
ekrengel: Maybe we should have this listed as a standard somewhere? I personally like `_ , __`

justinvp: The reason I used `_, __` is because otherwise the IDE complains:

```
'a' is declared but its value is never read. ts(6133)
```

The code fix suggestion is to prefix the arg names with an underscore, but I actually prefer _just_ using the underscore for unused arguments.

But happy to do whatever other folks prefer.

PR #172, opened 2020-01-06
Expose resource options

### LINKED ISSUE #110: Expose `ComponentResourceOptions` in policy rules
Policies should be able to access and check values from the `ComponentResourceOptions` object on resources. This would allow checking of `protect`, `additionalSecretOutputs`, and more for critical resource types.

justinvp:
Exposes resource options to policies.

This depends on https://github.com/pulumi/pulumi/pull/3720 (CI will fail until we have a new CLI release with those changes).

Here's a summary of resource options that we could potentially expose to policies:

```typescript
export interface ResourceOptions {
    id?: Input<ID>; // An optional existing ID to load, rather than create.
    parent?: Resource; // An optional parent resource to which this resource belongs.
    dependsOn?: Input<Input<Resource>[]> | Input<Resource>; // An optional additional explicit dependencies on other resources.
    protect?: boolean; // When set to true, protect ensures this resource cannot be deleted.
    ignoreChanges?: string[]; // Ignore changes to any of the specified properties.
    version?: string; // An optional version, corresponding to the version of the provider plugin that should be used when operating on this resource. This version overrides the version information inferred from the current package and should rarely be used.
    aliases?: Input<URN | Alias>[]; // An optional list of aliases to treat this resource as matching.
    provider?: ProviderResource; // An optional provider to use for this resource's CRUD operations.
    customTimeouts?: CustomTimeouts; // An optional customTimeouts configuration block.
    transformations?: ResourceTransformation[]; // Optional list of transformations to apply to this resource during construction.
}

export interface CustomTimeouts {
    create?: string; // The optional create timeout represented as a string e.g. 5m, 40s, 1d.
    update?: string; // The optional update timeout represented as a string e.g. 5m, 40s, 1d.
    delete?: string; // The optional delete timeout represented as a string e.g. 5m, 40s, 1d.
}

export interface CustomResourceOptions extends ResourceOptions {
    deleteBeforeReplace?: boolean; // When set to true, deleteBeforeReplace indicates that this resource should be deleted before its replacement is created when replacement is necessary.
    additionalSecretOutputs?: string[]; // The names of outputs for this resource that should be treated as secrets. This augments the list that the resource provider and pulumi engine already determine based on inputs to your resource. It can be used to mark certain ouputs as a secrets on a per resource basis.
    import?: ID; // When provided with a resource ID, import indicates that this resource's provider should import its state from the cloud resource with the given ID. The inputs to the resource's constructor must align with the resource's current state. Once a resource has been imported, the import property must be removed from the resource's options.
}

export interface ComponentResourceOptions extends ResourceOptions {
    providers?: Record<string, ProviderResource> | ProviderResource[]; // An optional set of providers to use for child resources. Either keyed by package name (e.g. "aws"), or just provided as an array.  In the latter case, the package name will be retrieved from the provider itself. In the case of a single provider, the options can be simplified to just pass along `provider: theProvider`. Note: do not provide both [provider] and [providers];
}
```

Here's how these are projected to policies. Please see my inline comments/questions:

```typescript
export interface PolicyResourceOptions {
    // Parent URN. For `validateStack` we could consider exposing this as `PolicyResource` instead
    // of as a string (URN) to make it more usable since we have all the resources available.
    // But I don't think we can for `validateResource` since we won't have data on all the
    // resources at this point.
    // For `validateStack`, we could also consider providing a helpful `children: PolicyResources[]` property
    // (or `getChildren` function) to make it easier for folks to inspect a resource's children without having
    // to traverse the resources and build-up the children array manually.
    parent?: string;

    protect: boolean;

    // URNs. Again, for `validateStack`, should we expose this as `PolicyResource[]` to make it easier to use
    // Or is it better to have a consistent signature with `validateResource`.
    dependencies: string[];

    // This could be an empty string for the stack resource and provider resources.
    // I implemented this before the holidays, but I am now wondering if we should make the
    // signature optional, e.g. `provider?: string`, instead of returning an empty string.
    // Any preference?
    provider: string;

    aliases: string[]; // URNs.
    customTimeouts?: PolicyCustomTimeouts;
    additionalSecretOutputs: string[];
}

// I explored making the timeout properties strings to match our Node.js SDK (e.g. "5m"),
// but I think it'll be much easier to inspect these in policies as numbers (of seconds)
// rather than as strings. These are just numbers inside the engine, so if we were to
// expose them as strings on the policy side, we'd have to normalize the string value,
// i.e. if someone wrote "1m" or "60s" in the Pulumi program, we'd have to pick one
// canonical form to expose from the policy SDK. It won't be super clear to policy authors
// which form they should be using inside their policies (i.e. should authors be looking for
// the string form that was specified in the Pulumi program or the normalized form?). Also,
// other language SDKs don't even expose these as strings but rather as idiomatic types for
// the language, such as `TimeSpan` in our .NET SDK.
export interface PolicyCustomTimeouts {
    create?: number;
    update?: number;
    delete?: number;
}
```

I chose not to expose the following properties for now:

```typescript
// I don't think it makes sense to expose this. When this is specified in a Pulumi program,
// the resource's state is read from the resource monitor instead of the resource being registered,
// and its state isn't part of the stack's state.
id?: Input<ID>; // An optional existing ID to load, rather than create.

// We don't have this information readily available for `validateStack`.
ignoreChanges?: string[]; // Ignore changes to any of the specified properties.

// This wasn't readily available, but I can investigate whether it's possible to easily expose this
// if there is a good scenario for inspecting it in policies. For now, I've left it out.
version?: string; // An optional version, corresponding to the version of the provider plugin that should be used when operating on this resource.

// This is purely something that happens inside the language SDKs and isn't available in a way that could
// be exposed in policies.
transformations?: ResourceTransformation[]; // Optional list of transformations to apply to this resource during construction.

// We don't have this information readily available for `validateStack`.
deleteBeforeReplace?: boolean; // When set to true, deleteBeforeReplace indicates that this resource should be deleted before its replacement is created when replacement is necessary.

// It's unclear to me how this would be used from policies. We can tell it's an import for `validateResource`,
// but would need to investigate how to expose for `validateStack`.
import?: ID; // When provided with a resource ID, import indicates that this resource's provider should import its state from the cloud resource with the given ID.
```

Fixes #110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ### Improvements
 
 - Add support for using `Config`, `getProject()`, `getStack()`, and `isDryRun()` from Policy Packs
-  via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 of the Pulumi SDK).
+  via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 or later of the Pulumi SDK) (https://github.com/pulumi/pulumi-policy/pull/169).
+
+- Expose resource options (https://github.com/pulumi/pulumi-policy/pull/172).
 
 ## 0.3.0 (2019-11-26)
 
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -157,9 +157,68 @@ export interface ResourceValidationArgs {
      */
     name: string;
 
-    // TODO: Add support for the following:
-    //
-    // opts: PolicyResourceOptions;
+    /**
+     * The options of the resource.
+     */
+    opts: PolicyResourceOptions;
+}
+
+/**
+ * PolicyResourceOptions is the bag of settings that control a resource's behavior.
+ */
+export interface PolicyResourceOptions {
+    /**
+     * An optional parent resource to which this resource belongs.
+     */
+    parent?: string;
+
+    /**
+     * When set to true, protect ensures this resource cannot be deleted.
+     */
+    protect: boolean;
+
+    /**
+     * Dependencies of this resource.
+     */
+    dependencies: string[];
+
+    /**
+     * The provider to use for this resource.
+     */
+    provider: string;
+
+    /**
+     * Additional URNs that should be aliased to this resource.
+     */
+    aliases: string[];
+
+    /**
+     * An optional customTimeouts configuration block.
+     */
+    customTimeouts?: PolicyCustomTimeouts;
+
+    /**
+     * Outputs that should always be treated as secrets.
+     */
+    additionalSecretOutputs: string[];
+}
+
+/**
+ * Optional custom timeout options.
+ */
+export interface PolicyCustomTimeouts {
+    /**
+     * The optional create timeout in seconds.
+     */
+    create?: number;
+    /**
+     * The optional update timeout in seconds.
+     */
+    update?: number;
+    /**
+     * The optional delete timeout in seconds.
+     */
+    delete?: number;
 }
 
 /**
@@ -241,9 +300,10 @@ export interface PolicyResource {
      */
     name: string;
 
-    // TODO: Add support for the following:
-    //
-    // opts: PolicyResourceOptions;
+    /**
+     * The options of the resource.
+     */
+    opts: PolicyResourceOptions;
 }
 
 /**
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -20,7 +20,9 @@ import { deserializeProperties } from "./deserialize";
 import {
     Policies,
     Policy,
+    PolicyCustomTimeouts,
     PolicyResource,
+    PolicyResourceOptions,
     ReportViolation,
     ResourceValidationArgs,
     ResourceValidationPolicy,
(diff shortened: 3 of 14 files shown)


justinvp:
Ahh, I thought I had accidentally added Pat as a reviewer (so I removed him). I didn't realize Luke added him. 🤦‍♂

ekrengel:
* Adding Children to Policy resource Options would be great!
* I think the Parents & dependencies should also be a Policy Resource → maybe we can use pointers or else this objects going to get huge. 
* Policy custom Timeouts → agree this would be best translated to Seconds. I am wondering if we should we make this clearer in the naming?
* Trying to wrap my head around the ignore changes. What happens if my program is passing pac but the resource is out of compliance due to a manual change. I think we probably want ignore changes to be able to prevent this from being used as a workaround, although I am also willing to punt on this until we have an ask.
* Version seems like it would be needed to properly check that a resource provider is the allowed version
* Delete before Replace seems like something I may want to write a policy against, although maybe we wait until we get the request.
* It's still a bit unclear to me why some info isn't available during 'validate stack' but is available for a resource validation. If this because a resource validation uses inputs while a validate stack uses outputs? Maybe we can use 2 separate struts for the resources opts then?

ekrengel:
@justinvp Whats going on with this PR and https://github.com/pulumi/pulumi/pull/3720? I am happy to approve 3720 although not sure I am the most suited to do so. I left you some comments/thoughts above, not sure if those got lost in the mist of holidays/planning!

justinvp:
@ekrengel, thanks for following-up on and apologies for not responding to your feedback sooner! I'm considering a richer object model for this and will post an update soon.

justinvp:
Closing in favor of #184

PR #173, opened 2020-01-06
Provide easier type checking for `validateStack`

### LINKED ISSUE #158: Provide easier type checking
_This may belong in @pulumi/pulumi but opening this in @pulumi/pulumi-policy since this Policy is the specific need/use case for it._

When doing stack validations you often (always?) need to check a resource's type. Currently, this involves something like this:
```ts
if (args.type === "azure:storage/account:Account") { /* ... */ }
```

We should provide an easier mechanism for doing this similar to our `isInstance(o)` method that doesn't require the user to know or learn the type's string value.

Related to #84.

justinvp:
- Deprecate `validateTypedResource` in favor of `validateResourceOfType`.
- Add `validateStackResourcesOfType` for strongly-type stack validations.
- Add `isType` and `asType` helper methods.

Usage:

```typescript
validateResource: validateResourceOfType(aws.s3.Bucket, (resource, args, reportViolation) => {
    // ...
});
```

```typescript
validateStack: validateStackResourcesOfType(aws.s3.Bucket, (resources, args, reportViolation) => {
    // ...
});
```

```typescript
if (resource.isType(aws.s3.Bucket) { ... }
```

```typescript
const bucket = resource.asType(aws.s3.Bucket);
if (bucket) { ... }
```

```typescript
const buckets = args.resources.map(r => r.asType(aws.s3.Bucket)).filter(b => b);
```

---
~This adds the following helper functions for use with `validateStack`:~
- ~`validateTypedResources` - similar to `validateResource`'s `validateTypedResource`.~
- ~`isTypedResource` - for checking if a `PolicyResource` is a resource type.~
- ~`asTypedResource` - for converting a `PolicyResource` to a strongly typed object.~
- ~`filterTypedResources` - for filtering an array of `PolicyResource`s to strongly-typed array.~

Fixes #158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Add support for using `Config`, `getProject()`, `getStack()`, and `isDryRun()` from Policy Packs
   via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 of the Pulumi SDK).
 
+- Provide easier type checking for `validateStack` (https://github.com/pulumi/pulumi-policy/pull/173).
+
 ## 0.3.0 (2019-11-26)
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,23 @@
 - Add support for using `Config`, `getProject()`, `getStack()`, and `isDryRun()` from Policy Packs
   via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 of the Pulumi SDK).
 
-- Provide easier type checking for `validateStack` (https://github.com/pulumi/pulumi-policy/pull/173).
+- Provide easier type checking for `validateStack`, along with `isType` and `asType` helper functions
+  (https://github.com/pulumi/pulumi-policy/pull/173).
+
+  Example:
+
+  ```typescript
+  {
+      validateStack: validateStackResourcesOfType(aws.s3.Bucket, (buckets, args, reportViolation) => {
+          for (const bucket of buckets) {
+              // ...
+          }
+      }),
+  }
+  ```
+
+- `validateTypedResource` is now deprecated in favor of `validateResourceOfType`. `validateTypedResource`
+  will be removed in an upcoming version. (https://github.com/pulumi/pulumi-policy/pull/173).
 
 ## 0.3.0 (2019-11-26)
 
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -160,6 +160,38 @@ export interface ResourceValidationArgs {
     // TODO: Add support for the following:
     //
     // opts: PolicyResourceOptions;
+
+    /**
+     * Returns true if the type of this resource is the same as `resourceClass`.
+     *
+     * For example:
+     *
+     * ```typescript
+     * if (args.isType(aws.s3.Bucket)) {
+     *     // ...
+     * }
+     * ```
+     */
+    isType<TResource extends Resource>(
+        resourceClass: { new(...rest: any[]): TResource },
+    ): boolean;
+
+    /**
+     * Returns the resource args for `resourceClass` if the type of this resource is the same as `resourceClass`,
+     * otherwise `undefined`.
+     *
+     * For example:
+     *
+     * ```typescript
+     * const bucketArgs = args.AsType(aws.s3.Bucket);
+     * if (bucketArgs) {
+     *     // ...
+     * }
+     * ```
+     */
+    asType<TResource extends Resource, TArgs>(
+        resourceClass: { new(name: string, args: TArgs, ...rest: any[]): TResource },
+    ): Unwrap<NonNullable<TArgs>> | undefined;
 }
 
 /**
@@ -176,14 +208,9 @@ export function validateTypedResource<TResource extends Resource, TArgs>(
         reportViolation: ReportViolation) => Promise<void> | void,
 ): ResourceValidation {
     return (args: ResourceValidationArgs, reportViolation: ReportViolation) => {
-        const isInstance = (<any>resourceClass).isInstance;
-        if (!isInstance || typeof isInstance !== "function") {
-            return;
+        if (args.isType(resourceClass)) {
+            validate(args.props as Unwrap<NonNullable<TArgs>>, args, reportViolation);
         }
-        if (isInstance({ __pulumiType: args.type }) !== true) {
-            return;
-        }
-        validate(args.props as Unwrap<NonNullable<TArgs>>, args, reportViolation);
     };
 }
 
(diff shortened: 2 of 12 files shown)


sdk/nodejs/policy/policy.ts
```diff
@@ -246,6 +246,72 @@ export interface PolicyResource {
...
+/**
+ * A helper function that returns a strongly-typed stack validation function.
+ * @param resourceClass A resource class used to filter this check to only resources of the specified class.
+ * @param validate A callback function that validates if a stack violates a policy.
+ */
+export function validateTypedResources<TResource extends Resource>(
+    resourceClass: { new(...rest: any[]): TResource },
+    validate: (
+        resources: q.ResolvedResource<TResource>[],
+        args: StackValidationArgs,
```
justinvp: Right now, the args is being passed as-is, but I am wondering if it'd be better to pass a modified `args` object with `args.resources` filtered to just the resources that match the type filter. The benefit of doing that would be you could then more easily inspect other metadata associated with the strongly-typed resources in `resources` via it's index in the array.

For example, say I'm using this new `validateTypedResources` method with `validateStack` to look at S3 buckets in a strongly-typed way, but also want to inspect those resource's options. If we provided a modified `args` with filtered `args.resources`, then inside the validation function you could do something like:

```typescript
validateStack: validateTypedResources(aws.s3.Bucket, (resources, args, reportViolation) => {
    for (let i = 0; i < resources.length; i++) {
        const bucket = resources[i];
        const bucketOptions = args.resources[i].options;
        ...
    }
}),
```

The downside is then you wouldn't have access to other types of resources from `args.resources` inside that validation function.

Alternatively, we don't pass a modified `args` (continue passing it unmodified) and if you want to look at other metadata associated with a resource, and still inspect the resource in a strongly-typed way, you don't use the `validateTypedResources` helper, but use some of the other helpers like:

```typescript
validateStack: (args, reportViolation) => {
    for (const r of args.resources)
        const bucket = asTypedResource(aws.s3.Bucket, r);
        if (!bucket) continue;
        const bucketOptions = r.options;
        ...
    }
},
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - Add support for using `Config`, `getProject()`, `getStack()`, and `isDryRun()` from Policy Packs
   via upgraded dependency on `@pulumi/pulumi` v1.8.0 (requires v1.8.0 of the Pulumi SDK).
 
+- Provide easier type checking for `validateStack` (https://github.com/pulumi/pulumi-policy/pull/173).
```
chrsmith: This is fine for now, but post our 1.0 launch we might want to have code snippets / examples in the PR's description so someone looking at the change log can find out more, etc. But while we are still feeling things out, the level of detail you have in this PR's description now is just fine.

sdk/nodejs/policy/policy.ts
chrsmith: Having more than one way to do something is generally an anti-pattern in software. So if we were to filter the `args.resources`, you'd be forced to use a different approach for writing the policy.

But TBH, I'm personally on-board with having N-different, very-specialized methods for defining policies. (e.g. have `validatedTypedResources` pre-filter `args.resources` to _just_ those that match the provided resource type.

The reason being is that all of these helper methods are designed to make it writing policies more expressive, and more declarative. A policy that can be expressed in 10 lines of code is more useful than the same functionality expressed in 20 lines of code, assuming there is no sacrifice in readability/clarity.

The only pushback I have though is that if we do have multiple ways to do something -- e.g. `validateTypedResources` and other similar helper methods -- we need to make them easier to use/mix and match. The name `validateTypedResources` doesn't really give an indication of what specifically is going on. But perhaps a different name like:

`validateStack: validateStackResourcesWithType(aws.s3.Bucket, (resources, args, reportViolation) => {`
or 
`validateStack: onlyResourcesWithType(aws.s3.Bucket, (resources, args, reportViolation) => {`
 etc.

Make sense?

ekrengel: I agree the naming could be more specific. I dont have a problem with the multiple helpers at least for now. I think we need to see how people are writing policies and what makes the most sense before making a call on whats the "best helper"

clstokes:
> This adds the following helper functions for use with `validateStack`:
> 
> * `validateTypedResources` - similar to `validateResource`'s `validateTypedResource`.
> * `isTypedResource` - for checking if a `PolicyResource` is a resource type.
> * `asTypedResource` - for converting a `PolicyResource` to a strongly typed object.
> * `filterTypedResources` - for filtering an array of `PolicyResource`s to strongly-typed array.
> 
> I'm open to better names for these.
> 
> Fixes #158

I'm good with these names and this covers what I was looking for in the original issue. LGTM

justinvp:
Thanks for the approval, Cameron! Though, I really love Chris's suggestions on the naming, particularly as it improves the readability of policies (thanks!).

My preference would be:

```typescript
validateStack: validateStackResourcesOfType(aws.s3.Bucket, (resources, args, reportViolation) => { ... });
```

And renaming the helper for `validateResource` from `validateTypedResource` to be along similar lines:

```typescript
validateResource: validateResourceOfType(aws.s3.Bucket, (resource, args, reportViolation) => { ... });
```

(We can keep the old `validateTypedResource` name as an alias marked as `@deprecrated` for now and remove it entirely in a subsequent release of the library, preferring the new name going forward.)

We could consider more terse naming, but I couldn't come up with any that felt quite right. If we wanted more terse names, we could do something like:

```typescript
validateStack: resourcesOfType(aws.s3.Bucket, (resources, args, reportViolation) => { ... });
// and
validateResource: ofType(aws.s3.Bucket, (resource, args, reportViolation) => { ... });
```

These read more nicely from left to right, e.g. "validate stack resources of type aws.s3.Bucket" and "validate resource of type aws.s3.Bucket", but don't stand as well on their own (i.e. it isn't as clear that `ofType` is a helper to be used with `validateResource`).

I'll plan on adopting the more verbose names, unless folks feel strongly about more terse ones.

---

While thinking about the naming here, I thought of a much better way to expose the other type helpers. Instead of providing these as top-level functions, they could be member functions of the objects themselves.

Instead of:

```typescript
if (isTypedResource(aws.s3.Bucket, resource)) { ... }
// and
const bucket = asTypedResource(aws.s3.Bucket, resource);
```

Something like:

```typescript
if (resource.isType(aws.s3.Bucket) { ... }
// and
const bucket = resource.asType(aws.s3.Bucket);
```

Having these as member functions are much more discoverable as you will see them in intellisense, rather than having to know about the top-level functions. And not having them as top-level functions makes the top-level functions we do have less confusing as they won't be lost among the more typically used `validateResourceOfType`/`validateStackResourcesOfType` functions.

A nice side benefit is that we can make the `asType` work with `validateResource`, where we want the resulting type to be the "input" shape (e.g. `BucketArgs`) rather than the "output" shape.

If we add these as member functions, I'd suggest holding off on adding a `filterTypedResources` equivalent for now as its functionality can be implemented quite simply and more idiomatically for JS/TS with the other more primitive member functions and array's built-in functions, e.g.:

```typescript
const buckets = args.resources
    .filter(r => r.isType(aws.s3.Bucket))
    .map(r => r.asType(aws.s3.Bucket));
// or
const buckets = args.resources
    .map(r => r.asType(aws.s3.Bucket))
    .filter(b => b);
```

And we could always add a convenience function that avoids having to use `map`/`filter` down the road, if necessary, based on actual feedback.

---

Based on above, I will add some more commits to this PR that does the following:

- Rename `validateTypedResource` (singular) to `validateResourceOfType` (with `validateTypedResource` still available but marked as `@deprecated`).
- Rename new `validateTypedResources` (plural) to `validateStackResourcesOfType`.
- Filter the `args.resources` in `validateStackResourcesOfType` so the strongly typed `resources` array and untyped `args.resources` array match.
- Delete new top-level `isTypedResource` and add `isType` member functions instead.
- Delete new top-level `asTypedResource` and add `asType` member functions instead.
- Delete new `filterTypedResources` for now.

justinvp:
@ekrengel, @clstokes, @chrsmith, I've added some more commits to this PR. PTAL. Thanks!

PR #179, opened 2020-01-29
Inline `asyncTest` utility function

justinvp:
We recently made a [change](https://github.com/pulumi/pulumi/pull/3809) in `@pulumi/pulumi` to mark this internal helper as `@internal` so that it would not be emitted in API docs. But TypeScript respects the `@internal` when attempting to use the API from an external package, so attempting to import `asyncTest` here no longer works. In this case, this utility code is simple enough to just copy over into this repository, rather than other workarounds to maintain using it as a dependency.


justinvp:
@ekrengel, @chrsmith, I'm going to merge this to unblock some other in-progress changes, but happy to address any feedback subsequently.

PR #180, opened 2020-01-30
Report violation when accessing unknowns during stack validation

### LINKED ISSUE #178: Unknown values accessed during stack validation do not throw as expected
We aren't wrapping each resource's `props` in a proxy object that throws for unknown values, like we do for resource validation functions.

justinvp:
For resource validations, an advisory violation is reported when accessing unknown values. We weren't doing this for stack validations.

This change fixes it for stack validations, and adds an integration test that ensures we're doing it for both resource and stack validations.

Fixes #178

### DIFF
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -43,7 +43,7 @@ func abortIfFailed(t *testing.T) {
 type policyTestScenario struct {
 	// WantErrors is the error message we expect to see in the command's output.
 	WantErrors []string
-	// Whether the error messages are advisory, and don't actually fail the command.
+	// Whether the error messages are advisory, and don't actually fail the operation.
 	Advisory bool
 }
 
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -43,6 +43,8 @@ func abortIfFailed(t *testing.T) {
 type policyTestScenario struct {
 	// WantErrors is the error message we expect to see in the command's output.
 	WantErrors []string
+	// Whether the error messages are advisory, and don't actually fail the command.
+	Advisory bool
 }
 
 // runPolicyPackIntegrationTest creates a new Pulumi stack and then runs through
@@ -147,7 +149,12 @@ func runPolicyPackIntegrationTest(
 				t.Log("No errors are expected.")
 				e.RunCommand(cmd, args...)
 			} else {
-				stdout, stderr := e.RunCommandExpectError(cmd, args...)
+				var stdout, stderr string
+				if scenario.Advisory {
+					stdout, stderr = e.RunCommand(cmd, args...)
+				} else {
+					stdout, stderr = e.RunCommandExpectError(cmd, args...)
+				}
 
 				for _, wantErr := range scenario.WantErrors {
 					inSTDOUT := strings.Contains(stdout, wantErr)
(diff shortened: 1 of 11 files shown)


tests/integration/integration_test.go
```diff
@@ -43,6 +43,8 @@ func abortIfFailed(t *testing.T) {
 type policyTestScenario struct {
 	// WantErrors is the error message we expect to see in the command's output.
 	WantErrors []string
+	// Whether the error messages are advisory, and don't actually fail the command.
```
chrsmith: Nit: Is "command" the best term to use here? Would "operation" be better? IMHO, it seems like a more natural fit. What do you think?

justinvp:
> Assuming I have that all right, here's the thing that doesn't quite sit right with me.

@chrsmith, I am glad you raised this, because the current behavior doesn't sit right with me either. This PR makes stack validations behave the same as resource validations, but I was not part of the original design for it always resulting in an `advisory` violation. I had the same thoughts about, "but what if I actually want to treat these as `mandatory` violations (i.e. "treat warnings as errors").

Instead of always emitting these as `advisory`, would it make more sense to always emit them as the configured enforcement level for the policy? So if my policy is configured to be `mandatory`, they would be emitted as `mandatory` as well?

justinvp:
@ekrengel, do you have any context on the current behavior?

chrsmith:
Happy to defer to you and Erin on the design here, but here's my opinion:

> Instead of always emitting these as advisory, would it make more sense to always emit them as the configured enforcement level for the policy?

I don't think that's the right way to go. Instead, I think we'd have some other error level like "information" or "warning" and use _that_.

Because this particular class of violation -- "accessing an output/computed/unknown value" -- isn't really a policy violation at all. It's just an indication that the policy you want to run _couldn't_. And unfortunately, there isn't anything we can do about it.

If we were to emit a "mandatory" violation, and fail the preview in these cases, the end user would be unable to use PAC. Since it could just happen that resource X depends on the output of resource Y. And so it would be lame for that particular stack to be prevented from updating because of the organization's policy group settings. (Since it isn't that there is a problem with the stack, just a lack of confirmation.)

Does that make sense?

The thing I personally would like to avoid is having these unpreventable policy violations essentially prevent people from being able to use PAC in for everyday use. And so while it's important we call these situations out, I think we should do so in a way that doesn't cause any undo friction. (Though perhaps we could offer a "treat-info/warning-violations-as-errors" setting too... so that we would prevent stack previews if they ever needed to access an output property, etc.)

justinvp:
> I don't think that's the right way to go. Instead, I think we'd have some other error level like "information" or "warning" and use _that_.

This makes sense and I agree. Though, I am inclined to still move forward with this PR to make `validateStack` behave the same as `validateResource`, and then open a separate issue to track changing both to using some other type of info/warning level.

ekrengel:
> Though, I am inclined to still move forward with this PR to make validateStack behave the same as validateResource, and then open a separate issue to track changing both to using some other type of info/warning level.

I would agree this would be the correct was to proceed. 

I was not super involved in the discussions / decisions of what behavior to take when accessing outputs BUT I believe @lukehoban and @pgavlin were.

justinvp:
Opened #181 to track changing the behavior here

PR #205, opened 2020-02-29
Add `enforcementLevel` to `PolicyPackArgs`

### LINKED ISSUE #192: Add top-level `enforcementLevel` to `PolicyPackArgs`
This is just a convenience for policy pack authors. Instead of having to define a default `enforcementLevel` value for each policy in the policy pack, this provides a place where such a default can be specified once (and setting it on any individual policies would then override the default value for those policies).

```typescript
// Existing interface.
    policies: Policies;

    // New property.
}
```

And change the `enforcementLevel` on the existing `Policy` to be optional.

justinvp:
This is just a convenience for policy pack authors. Instead of having to define a default `enforcementLevel` value for each policy in the policy pack, this provides a place where such a default can be specified once. Setting it on any individual policies overrides the default value for those policies. If no `enforcementLevel` is set, defaults to "advisory".

~~Testing this given the current implementation isn't super easy. It requires modularizing the implementation so it can be more easily unit tested.~~

Edit: Added integration tests

Fixes #192

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -16,6 +16,8 @@ import { Resource, Unwrap } from "@pulumi/pulumi";
 import * as q from "@pulumi/pulumi/queryable";
 import { serve } from "./server";
 
+const defaultEnforcementLevel: EnforcementLevel = "advisory";
+
 /**
  * The set of arguments for constructing a PolicyPack.
  */
@@ -24,6 +26,13 @@ export interface PolicyPackArgs {
      * The policies associated with a PolicyPack.
      */
     policies: Policies;
+
+    /**
+     * Indicates what to do on policy violation, e.g., block deployment but allow override with
+     * proper permissions. Default for all policies in the PolicyPack. Individual policies can
+     * override.
+     */
+    enforcementLevel?: EnforcementLevel;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


sdk/nodejs/policy/policy.ts
```diff
@@ -24,6 +26,13 @@ export interface PolicyPackArgs {
      * The policies associated with a PolicyPack.
      */
     policies: Policies;
+
+    /**
+     * Indicates what to do on policy violation, e.g., block deployment but allow override with
+     * proper permissions. Default for all policies in the PolicyPack. Individual policies can
+     * override.
+     */
+    enforcementLevel?: EnforcementLevel;
```
ekrengel: Hmm I am wondering if from a service perspective we should have the ability to set "undefined" (which wouldnt really be undefined),  but basically a way to say default to the all enforcement level instead of using the policy specific one.

sdk/nodejs/policy/server.ts
```diff
@@ -87,9 +94,9 @@ export function serve(policyPackName: string, policyPackVersion: string, policie
     // Finally connect up the gRPC client/server and listen for incoming requests.
     const server = new grpc.Server();
     server.addService(analyzerrpc.AnalyzerService, {
-        analyze: makeAnalyzeRpcFun(policyPackName, policyPackVersion, policies),
-        analyzeStack: makeAnalyzeStackRpcFun(policyPackName, policyPackVersion, policies),
-        getAnalyzerInfo: makeGetAnalyzerInfoRpcFun(policyPackName, policyPackVersion, policies),
+        analyze: makeAnalyzeRpcFun(policyPackName, policyPackVersion, policyPackEnforcementLevel, policies),
```
ekrengel: Hmm instead of constantly changing this interface, I am wondering if it makes sense to make a PolicyPack type 🤔

PR #206, opened 2020-03-01
Use latest released CLI/package and update tests

justinvp:
Update dependencies to no longer depend on dev releases, and update integration tests to expect the new violation output format.


justinvp:
@ekrengel, @chrsmith, I'm going to merge this "To Be Reviewed," as I'm preparing other changes with tests that make use of the new violation output format. Happy to address any feedback subsequently.

Issue #188, opened 2020-02-06
Config: Add object model API for accessing config from policy validation handlers


justinvp:
Add object model API for accessing config from policy validation handlers in `@pulumi/policy`.

This involves adding the a method on the `args` interface for `validateResource` and `validateStack`:

```typescript
getConfig<T>(): T;
```

Issue #187, opened 2020-02-06
Config: Add object model API to declare schema


justinvp:
Add object model API to declare schema in `@pulumi/policy`.

Basically this will be a set of TypeScript interfaces that describe the subset of JSON schema we want to support when specifying the config schema when declaring policies.

Something like:

```typescript
export interface Policy {
    // New property on existing interface.
    config?: PolicyConfigSchema;
}

// New interface.
export interface PolicyConfigSchema {
    properties: Record<string, JSONSchema>;
    required?: string[];
}

// New interface.
export interface JSONSchema {
    ...
}
```

For JSONSchema consider using https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/json-schema or just extracting a subset of it.

Issue #190, opened 2020-02-06
Config: Use config from the service for required policy packs


justinvp:
Pass the config from the service to the policy pack when running it. The service returns a list of required policy packs. Any config will be returned as part of that for each policy pack. We need to pass this config along to each policy pack.

Issue #189, opened 2020-02-06
Config: Send config schema to service


justinvp:
- Return schema as part of policy pack's metadata
- Send to service as part of sending other metadata

This involves returning the schema information from the `GetAnalyzerInfo` gRPC method.

When sending the schema to the service, do we synthesize a single schema for the entire Policy Pack, or send separate smaller JSON schema's for each config property on each policy?

PR #207, opened 2020-03-03
Initial support for configuring policies

justinvp:
Broken up by commits for easier reviewing of this larger PR. I'm still finishing writing several integration tests (which may require minor bug fixes), but there's no reason this can't be reviewed in the meantime.

The CLI side of the change is here: https://github.com/pulumi/pulumi/pull/4015. This PR will fail until that is merged and we can take a dependency on it.

Fixes https://github.com/pulumi/pulumi-policy/issues/187
Fixes https://github.com/pulumi/pulumi-policy/issues/188

Part of https://github.com/pulumi/pulumi-policy/issues/185
Part of https://github.com/pulumi/pulumi-policy/issues/186
Part of https://github.com/pulumi/pulumi-policy/issues/189
Part of https://github.com/pulumi/pulumi-policy/issues/190


sdk/nodejs/policy/policy.ts
```diff
@@ -102,8 +102,67 @@ export interface Policy {
...
+    dependencies?: {
+        [key: string]: PolicyConfigPropertySchemaDefinition | string[];
+    };
+    propertyNames?: PolicyConfigPropertySchemaDefinition;
+
+    description?: string;
+    default?: PolicyConfigPropertySchemaType;
+    readOnly?: boolean;
+    writeOnly?: boolean;
+}
```
justinvp: It's not clear to me we want to expose all this, at least not initially. Should we just expose a minimal useful subset and then see if there's feedback before exposing more? If so, what would the subset look like? I just added most of it from [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a9a4d049259d3748c2d1440238489cbccf3cd395/types/json-schema/index.d.ts#L545) as a stand-in.

sdk/nodejs/policy/policy.ts
```diff
@@ -102,8 +102,67 @@ export interface Policy {
      * proper permissions.
      */
     enforcementLevel?: EnforcementLevel;
+
+    /**
+     * Configuration schema.
+     */
+    config?: PolicyConfigSchema;
```
justinvp: I keep going back and forth between calling this `config` vs. `schema` vs. `configSchema`. Any preferences? The latter is the most clear (though, slightly more verbose). [ESLint calls its equivalent `schema`](https://eslint.org/docs/developer-guide/working-with-rules#options-schemas) whereas [TSLint calls its equivalent `options`](https://github.com/palantir/tslint/blob/5ba66591baee97d0514c27bb106b1e4b62c7853e/src/rules/noAnyRule.ts#L47-L52).

ekrengel: I prefer configSchema for the schema.

sdk/nodejs/policy/policy.ts
ekrengel: I am not sure `readOnly` and `writeOnly` make sense here. I think my preference would be to start with a smaller batch of well tested ones and then grow it with time. I think for starters we should support: type, enum, all the num validation ones, all the string validation ones, and all the list validation ones, all the property validation ones... and is i begin to compile this list I ask may be we just support them all 🤔

sdk/nodejs/policy/server.ts
```diff
@@ -157,7 +193,12 @@ function makeAnalyzeRpcFun(
         const ds: Diagnostic[] = [];
         try {
             for (const p of policies) {
-                const enforcementLevel = p.enforcementLevel || policyPackEnforcementLevel;
+                // If there is a configured enforcement level for the policy, it is used. Otherwise, the enforcement
```
ekrengel: Im not quite sure what this comment is saying! (as well as below on line 323, which appears to be the same comment)

sdk/nodejs/policy/policy.ts
justinvp: Here's what I've done:

- Removed `readOnly` and `writeOnly`.
- Added `format`.
- Kept the rest that I had.

Here's all of the properties (json schema 7) that I've left off for now:

```typescript
    $id?: string;
    $ref?: string;
    $schema?: JSONSchema7Version;
    $comment?: string;

    if?: JSONSchema7Definition;
    then?: JSONSchema7Definition;
    else?: JSONSchema7Definition;

    allOf?: JSONSchema7Definition[];
    anyOf?: JSONSchema7Definition[];
    oneOf?: JSONSchema7Definition[];
    not?: JSONSchema7Definition;

    contentMediaType?: string;
    contentEncoding?: string;

    definitions?: {
        [key: string]: JSONSchema7Definition;
    };

    title?: string;
    readOnly?: boolean;
    writeOnly?: boolean;
    examples?: JSONSchema7Type;
```

I also decided to move most of the schema types out of the main module, to avoid the unnecessary noise when importing via `import * as policy from "@pulumi/policy"`.

And I added some initial doc comments, though, as a follow-up, I'd like to add doc comments to all of these.

sdk/nodejs/policy/policy.ts
justinvp: Changed to `configSchema`.

sdk/nodejs/policy/server.ts
justinvp: Decide to just remove the comment instead of rewording. From the code, it should be clear where we're getting the enforcementLevel value.

justinvp:
Added more commits that addresses the feedback (see commit messages) and adds some more integration tests.

Issue #208, opened 2020-03-10
Basic support for writing policies in Python


justinvp:
- [x] (PR https://github.com/pulumi/pulumi/pull/4057) Changes in `pulumi/pulumi` to launch Python Policy Packs
- [x] (PR https://github.com/pulumi/pulumi-policy/pull/212) Python SDK library in `pulumi/pulumi-policy` for creating policies
- [x] (PR https://github.com/pulumi/pulumi-policy/pull/212) Hook Python library to build
- [x] (PR https://github.com/pulumi/pulumi-policy/pull/212) Run Python policies in integration tests
- [x] (PR https://github.com/pulumi/pulumi-policy/pull/212) Expose resource dependencies
- [x] (PR https://github.com/pulumi/pulumi-policy/pull/216) Deserialize properties fully (e.g. Assets/Archives are deserialized to appropriate types)
- [ ] (PR https://github.com/pulumi/pulumi-policy/pull/218) Unknown checking proxy for properties (https://github.com/pulumi/pulumi-policy/issues/209)
- [x] Automate publishing SDK library to PyPi
- [x] (PR https://github.com/pulumi/templates-policy/pull/16) Add Python policy templates

justinvp:
Going to close this issue as done. There's one remaining task that's tracked in its own issue #209 that has a PR out #218.

PR #212, opened 2020-03-13
Support for writing policies in Python

justinvp:
Adds basic support for writing policies in Python.

Depends on https://github.com/pulumi/pulumi/pull/4057 (will fail until we have a new dev release of the CLI along with new dev release of the `pulumi` PyPi package that we can take a dependency on).

There are some remaining TODOs noted in comments, that I'll address in follow-up changes, tracked as tasks in #208.

Part of #208


sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -0,0 +1,707 @@
+# Copyright 2016-2020, Pulumi Corporation.
```
justinvp: Note: I started breaking up this file into smaller files, but ran into circular imports that I need to chase down, so keeping it all in one file for the time-being.

tests/integration/validate_resource/policy-pack-python/__main__.py
```diff
@@ -0,0 +1,76 @@
...
+    if args.resource_type == "pulumi-nodejs:dynamic:Resource":
+        if "state" in args.props and args.props["state"] == 3:
+            report_violation("'state' must not have the value 3.")
+
+def dynamic_no_state_with_value_4(args: ResourceValidationArgs, report_violation: ReportViolation):
+    if args.resource_type == "pulumi-nodejs:dynamic:Resource":
+        if "state" in args.props and args.props["state"] == 4:
+            report_violation("'state' must not have the value 4.")
+
+# Note: In the NodeJS Policy Pack, this is a strongly-typed policy, but since Python
```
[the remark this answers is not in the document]
justinvp: Not for GA. I have some ideas how it _could_ work for Python, though.

sdk/python/lib/pulumi_policy/version.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pkg_resources
+
+def _get_sem_version(package_version: str) -> str:
```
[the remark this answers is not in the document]
justinvp: This isn't actually sent to the service. It's the version of the `pulumi_policy` package and is used to satisfy this rpc: https://github.com/pulumi/pulumi/blob/cfcfad2f3c8c4bb7f34130f68e2c842dda3dc7e6/sdk/proto/analyzer.proto#L38

Here's the TypeScript equivalent:

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/sdk/nodejs/policy/server.ts#L149-L157

Which uses this imported version:

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/sdk/nodejs/policy/server.ts#L46

Which is populated as part of the build process:

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/sdk/nodejs/policy/version.ts#L15-L16

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/sdk/nodejs/policy/Makefile#L17

sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -0,0 +1,751 @@
...
+            name, version, policies, enforcement_level if enforcement_level is not None else EnforcementLevel.ADVISORY)
+        server = grpc.server(futures.ThreadPoolExecutor(max_workers=4))
+        analyzer_pb2_grpc.add_AnalyzerServicer_to_server(
+            servicer, server)
+        port = server.add_insecure_port(address="0.0.0.0:0")
+        server.start()
+        sys.stdout.buffer.write(f"{port}\n".encode())
+        try:
+            while True:
+                time.sleep(_ONE_DAY_IN_SECONDS)
```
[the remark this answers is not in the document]
justinvp: I based this on how we do it for Python dynamic providers. See https://github.com/pulumi/pulumi/blob/cfcfad2f3c8c4bb7f34130f68e2c842dda3dc7e6/sdk/python/lib/pulumi/dynamic/__main__.py#L169-L180

sdk/python/lib/pulumi_policy/version.py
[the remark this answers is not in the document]
justinvp: Note: When we build the Python package, we start with the version we'd use for NodeJS and then convert that to a PEP-440 compliant version.

https://github.com/pulumi/pulumi-policy/pull/212/files#diff-3372d3d66635fc93bc6dc20d49b825a6R2

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/scripts/get-py-version#L1-L7

https://github.com/pulumi/pulumi-policy/blob/e830f855f46c7bfbe77bbfc359b78a1ef0626acd/scripts/get-py-version.go#L1-L23

Which calls the following in `pulumi/pulumi`:

https://github.com/pulumi/pulumi/blob/2b3c7f61c54bc550ad79c6a7f6ee0cbd4192ceb4/pkg/util/buildutil/semver.go

https://github.com/pulumi/pulumi/blob/2b3c7f61c54bc550ad79c6a7f6ee0cbd4192ceb4/pkg/util/buildutil/semver_test.go

This code here effectively does the opposite, translating the PEP-440 compliant version back to a semver.

CHANGELOG.md
```diff
@@ -38,6 +38,31 @@
...
+      name="aws-policy-pack",
+      enforcement_level=EnforcementLevel.MANDATORY,
+      policies=[
+          ResourceValidationPolicy(
+              name="s3-no-public-read",
+              description="Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
+              validate=s3_no_public_read,
+          ),
+      ],
+  )
```
justinvp: Worth mentioning: there are two ways to define policies in Python that Luke and I discussed.

First is this way, which is similar to how policies are declared in TypeScript. Unfortunately, lambdas in Python must be expressions and cannot have statements, so it necessitates defining the handler function separately.

The second way involves inheriting from `ResourceValidationPolicy` or `StackValidationPolicy` and overriding the `validate` function. This would look like:

```python
class S3NoPublicRead(ResourceValidationPolicy):
    def __init__(self):
        super().__init__(
            name="s3-no-public-read",
            description="Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.")

    def validate(args, report_violation):
        if args.resource_type == "aws:s3/bucket:Bucket" and "acl" in args.props:
              acl = args.props["acl"]
              if acl == "public-read" or acl == "public-read-write":
                  report_violation("You cannot set public-read or public-read-write on an S3 bucket.")

PolicyPack(
      name="aws-policy-pack",
      enforcement_level=EnforcementLevel.MANDATORY,
      policies=[
          S3NoPublicRead(),
      ],
  )
```

Luke and I also chatted about possibly using Python decorators to automatically add policies with a certain decorator to a Policy Pack, but that is not implemented in this PR (but could potentially be built on-top of this lower-level support). I have not yet had a chance to experiment with this.

sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -0,0 +1,751 @@
...
+urn of the resource to associate with the violations.
+"""
+
+
+class ResourceValidationArgs:
+    """
+    ResourceValidationArgs is the argument bag passed to a resource validation.
+    """
+
+    resource_type: str
```
justinvp: Ideally, this would be `type`, but `type` is reserved, so we could either go with `type_` or something more meaningful like `resource_type`. Luke and I chatted about this and decided to go with the latter, even though that's different than what we've done with resource transformations [here](https://github.com/pulumi/pulumi/blob/cfcfad2f3c8c4bb7f34130f68e2c842dda3dc7e6/sdk/python/lib/pulumi/resource.py#L222). (We probably should have used `resource_type` for transformations.)

sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -0,0 +1,751 @@
...
+        policies: List[proto.PolicyInfo] = []
+        for policy in self.__policies:
+            enforcement_level = (policy.enforcement_level if policy.enforcement_level is not None
+                                 else self.__policy_pack_enforcement_level)
+            policies.append(proto.PolicyInfo(**{
+                "name": policy.name,
+                "description": policy.description,
+                "enforcementLevel": self._map_enforcement_level(enforcement_level),
+                # TODO[pulumi/pulumi-policy#210]: Expose config schema
+            }))
```
lukehoban: Since we know the names, can't we just pass as normal named params `proto.PolicyInfo(name=policy.name, ...)`?  Same for all the other calls here.

sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -0,0 +1,751 @@
...
+            enforcement_level = self._get_enforcement_level(policy)
+            if enforcement_level == EnforcementLevel.DISABLED or not isinstance(policy, ResourceValidationPolicy):
+                continue
+
+            report_violation = self._create_report_violation(diagnostics, policy.name,
+                                                             policy.description, enforcement_level)
+
+            # TODO[pulumi/pulumi-policy#208]: Deserialize properties
+            # TODO[pulumi/pulumi-policy#208]: Unknown checking proxy
+            props = json_format.MessageToDict(request.properties)
```
lukehoban: What does happen to "unknowns" in the meantime?

sdk/python/lib/pulumi_policy/version.py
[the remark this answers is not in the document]
lukehoban: It is unfortunate to need all this logic here - I would love if we could avoid some/most of it or centralize somehow.

sdk/python/lib/setup.py
```diff
@@ -0,0 +1,41 @@
...
+      version='${VERSION}',
+      description='Pulumi\'s Policy Python SDK',
+      long_description=readme(),
+      long_description_content_type='text/markdown',
+      url='https://github.com/pulumi/pulumi-policy',
+      license='Apache 2.0',
+      packages=find_packages(exclude=("test*",)),
+      package_data={
+          'pulumi_policy': [
+              'py.typed'
```
lukehoban: Do we have a `py.typed`?  I believe it can just be an empty file we place at this location.

sdk/python/lib/pulumi_policy/policy.py
justinvp: They'll have unknown sentinel values.

```go
const (
	// UnknownBoolValue is a sentinel indicating that a bool property's value is not known, because it depends on
	// a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownBoolValue = "1c4a061d-8072-4f0a-a4cb-0ff528b18fe7"
	// UnknownNumberValue is a sentinel indicating that a number property's value is not known, because it depends on
	// a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownNumberValue = "3eeb2bf0-c639-47a8-9e75-3b44932eb421"
	// UnknownStringValue is a sentinel indicating that a string property's value is not known, because it depends on
	// a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownStringValue = "04da6b54-80e4-46f7-96ec-b56ff0331ba9"
	// UnknownArrayValue is a sentinel indicating that an array property's value is not known, because it depends on
	// a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownArrayValue = "6a19a0b0-7e62-4c92-b797-7f8e31da9cc2"
	// UnknownAssetValue is a sentinel indicating that an asset property's value is not known, because it depends on
	// a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownAssetValue = "030794c1-ac77-496b-92df-f27374a8bd58"
	// UnknownArchiveValue is a sentinel indicating that an archive property's value is not known, because it depends
	// on a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownArchiveValue = "e48ece36-62e2-4504-bad9-02848725956a"
	// UnknownObjectValue is a sentinel indicating that an archive property's value is not known, because it depends
	// on a computation with values whose values themselves are not yet known (e.g., dependent upon an output property).
	UnknownObjectValue = "dd056dcd-154b-4c76-9bd3-c8f88648b5ff"
)
```

Planning to quickly address this, so it shouldn't be a problem for long. Aiming to have the change ready before the hackathon.

sdk/python/lib/pulumi_policy/version.py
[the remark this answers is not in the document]
justinvp: I've changed this to just emit the semver directly into the source code during build, so we can remove all this logic.

sdk/python/lib/pulumi_policy/policy.py
justinvp: Yep. Fixed.

sdk/python/lib/setup.py
justinvp: Added.

PR #218, opened 2020-03-25
Add wrappers for dict/list that raise on unknown values

### LINKED ISSUE #209: Python: Warn when accessing unknown values
Create Python proxy/wrapper around properties that throw for unknown values.

justinvp:
This allows us to skip running a policy when it accesses an unknown value. When that happens, and advisory diagnostic is emitted. (Note: as a future improvement, we plan to create a new kind of warning event for these cases, as this isn't really an advisory violation -- this just matches the current behavior we have for Node.js policies.)

Fixes #209


ekrengel:
> Note: This PR builds on-top of PR #216, so the first two commits can be ignored. 

When do you expect that to be merged? I tried viewing but its a bit hard to follow right now.

sdk/python/lib/pulumi_policy/policy.py
```diff
@@ -631,11 +643,22 @@ def AnalyzeStack(self, request, context):
...
+                if isawaitable(result):
+                    loop = asyncio.new_event_loop()
+                    loop.run_until_complete(result)
+                    loop.close()
+            except UnknownValueError as e:
+                diagnostics.append(proto.AnalyzeDiagnostic(
+                    policyName=policy.name,
+                    policyPackName=self.__policy_pack_name,
+                    policyPackVersion=self.__policy_pack_version,
+                    message=f"can't run policy '{policy.name}' during preview: {e.message}",
```
ekrengel: What does this look like when printed? Wondering if this should be more descriptive?

lukehoban: I wondered the same actually.  Note that this is the same as TS though:

https://github.com/pulumi/pulumi-policy/blob/5f41c47c0945414d142e149745a88cd6b9334f64/sdk/nodejs/policy/server.ts#L269

justinvp:
> Is this already done for TypeScript?

Yes, the Node.js implementation is at https://github.com/pulumi/pulumi-policy/blob/master/sdk/nodejs/policy/proxy.ts

> What does this look like when printed?

The output is the same as the existing Node.js output. From the integration test:

https://github.com/pulumi/pulumi-policy/blob/5f41c47c0945414d142e149745a88cd6b9334f64/tests/integration/integration_test.go#L431-L432

It keeps track of the "path", so if you're accessing a deeply nested unknown value like: `resource.props.foo.bar.baz` and `baz` is unknown, the message will say "value at .foo.bar.baz can't be known during preview"

PR #228, opened 2020-04-08
Update README.md and prepare for 1.0

justinvp:


justinvp:
@ekrengel, I simplified the README to reduce the duplicated content we have with the docs, in favor of just linking to the docs.

Issue #233, opened 2020-04-09
Replace `grpc` with `@grpc/grpc-js`


justinvp:

justinvp:
WIP: https://github.com/pulumi/pulumi-policy/pull/235

PR #235, opened 2020-04-13
Update dependencies to 2.0

justinvp:
As part of this, replace `grpc` with `@grpc/grpc-js` to match `@pulumi/pulumi` 2.0.

Fixes https://github.com/pulumi/pulumi-policy/issues/233
Part of https://github.com/pulumi/pulumi-policy/issues/236


justinvp:
The change from the native `grcp` to pure JavaScript `@grpc/grpc-js` is to match what we've done in `@pulumi/pulumi` in https://github.com/pulumi/pulumi/pull/3728. It means the released version of `@pulumi/policy` that includes this change will only work with Pulumi CLI >= 1.14.0 (because this is needed https://github.com/pulumi/pulumi/pull/4253).

justinvp:
After this lands, I'm planning to prepare a v1.1.0 release of `@pulumi/policy`. Or should it be `v1.0.1`?

PR #275, opened 2021-10-27
Migrating from travis to GHA

emiliza:
Experimental run with the remote component tests in https://github.com/pulumi/pulumi-policy/pull/274 with these changes in: https://github.com/pulumi/pulumi-policy/runs/4268990932?check_suite_focus=true

Haven't tested the master and release workflows


.github/workflows/run-acceptance-tests.yml
```diff
@@ -0,0 +1,127 @@
...
+        uses: actions/checkout@v2
+        with:
+          path: ci-scripts
+          repository: pulumi/scripts
+      - name: Build SDK
+        run: make only_build
+      - name: Check worktree clean
+        run: ./ci-scripts/ci/check-worktree-is-clean
+      - name: Run Tests
+        run: make only_test_fast
```
justinvp, line 113: This will run each library's unit tests. We're also going to want to run the integration tests:

https://github.com/pulumi/pulumi-policy/blob/be2c236f8ece50888b0686f245bbd4e071a2a6b8/Makefile#L21-L22

justinvp (commented):
Note that the README.md has a Travis badge that we'll need to change for GHA.

sdk/python/Makefile
```diff
@@ -14,8 +14,8 @@ ensure::
 
 build::
 	rm -rf $(PYENVSRC) && cp -R ./lib/. $(PYENVSRC)/
-	sed -i.bak "s/\$${VERSION}/$(VERSION)/g" $(PYENVSRC)/setup.py && rm $(PYENVSRC)/setup.py.bak
-	sed -i.bak "s/\$${SEMVERSION}/$(SEMVER:v%=%)/g" $(PYENVSRCLIB)/version.py && rm $(PYENVSRCLIB)/version.py.bak
```
justinvp, line 18: From the test failures, looks like we need to keep this line.

justinvp (approved):
LGTM. Let's merge it!

PR #280, opened 2021-11-23
fix for ci-scripts and cleanup

emiliza:
Fixes: 
```
$(go env GOPATH)/src/github.com/pulumi/scripts/ci/build-package-docs.sh policy
/bin/bash: /home/runner/go/src/github.com/pulumi/scripts/ci/build-package-docs.sh: No such file or directory```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,19 +11,8 @@ ensure::
 publish_packages:
 	$(call STEP_MESSAGE)
 	./scripts/publish_packages.sh
-	$$(go env GOPATH)/src/github.com/pulumi/scripts/ci/build-package-docs.sh ${PROJECT_NAME}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -1,5 +1,3 @@
-PROJECT_NAME := policy
-SUB_PROJECTS := sdk/nodejs/policy sdk/python
```
justinvp, line 2: I think we need to keep these two lines for `build/common.mk`.

justinvp (approved):
LGTM otherwise

Makefile
emiliza: Whoops, got overzealous in cleaning there.

PR #281, opened 2021-11-23
set env variables for docs build

emiliza:
At some point we should revisit updating [the script](https://github.com/pulumi/scripts/blob/master/ci/build-package-docs.sh), but for now these vars should get it through...

Apologies for this is turning out to be such a hot potato @justinvp  - not really sure how best to test this otherwise


justinvp (approved):
> Apologies for this is turning out to be such a hot potato

No problem. I don't think there's really a better way to try out the changes.

PR #287, opened 2022-09-28
Use loopback addresses in gRPC server

justinvp:
Improves security and should avoid tripping Windows & macOS firewall when running policies.


justinvp:
Will rebase after #288 is merged.

Frassle (approved):

PR #314, opened 2023-10-01
Policy remediations SDK support

joeduffy:
**Note**: This is still a work in progress. I will tag reviewers once ready.

This PR implements the TypeScript and Python policy SDK side of support for the new policy remediations feature. This lets developers implement new remediations callbacks, similar to policies, in their policy packs. See https://github.com/pulumi/pulumi/pull/14080 for the engine/CLI support, which is a prerequisite for this working.


joeduffy:
@justinvp This is fairly close. Probably good time to take an early look. I am mostly writing tests at this point.

sdk/python/lib/pulumi_policy/serialize.py
```diff
@@ -85,3 +87,49 @@ def _deserialize_archive(prop: Dict[str, Any]) -> pulumi.Archive:
...
+    return result
+
+def _serialize_property(prop: Any) -> Any:
+    # Check for proxies, and unwrap to get the raw elements. This ensures unknowns flow through.
+    if isinstance(prop, _DictProxy):
+        return _serialize_property(prop["__map"])
+    if isinstance(prop, _ListProxy):
+        return _serialize_property(prop["__elems"])
+
+    if isinstance(prop, list):
```
justinvp, line 107: We could make this more general and check for `Sequence` rather than `list`... But that has its own oddities that make it a bit more complex, where you have to explicitly check for `str` and some other things that are also `Sequence` that we don't want to treat as a `Sequence` (see https://github.com/pulumi/pulumi/blob/a935a23a41e5806c5cdeea8cf7ea4dd869947058/sdk/python/lib/pulumi/runtime/rpc.py#L330-L332), so I'd understand wanting to keep this check `isinstance(prop, list)`.

sdk/nodejs/policy/server.ts
```diff
@@ -573,3 +586,102 @@ function isTypeOf<TResource extends Resource>(
...
+                    if (result && result.getOriginalTarget) {
+                        result = result.getOriginalTarget();
+                    }
+
+                    rs.push({
+                        policyName: p.name,
+                        policyPackName,
+                        policyPackVersion,
+                        description: p.description,
+                        properties: result,
```
justinvp, line 674: Don't we need to serialize the result? Turn assets/archives back into special objects, etc.?

justinvp (commented):
What are we going to do about secrets? I haven’t tried it, but it looks like we’re going to lose those in a remediation, because when we deserialize the incoming properties here in the policy SDK, we strip the secretness of values, and then when we return the resulting properties, any secrets that were present will no longer be present going out. Perhaps we need a proxy object that can keep track of secrets?

Separate from that, we may also want to provide a way for a remediation to introduce a new value that’s a secret. Maybe we could allow users to use `pulumi.secret`/`Output.secret` from the core SDKs when, and update the serialization code here to support serializing such outputs?

sdk/nodejs/policy/policy.ts
```diff
@@ -25,9 +25,9 @@ const defaultEnforcementLevel: EnforcementLevel = "advisory";
  */
 export interface PolicyPackArgs {
     /**
-     * The policies associated with a PolicyPack.
+     * The policies associated with a PolicyPack. These check for and enforce policies.
      */
-    policies: Policies;
+    policies?: Policies;
```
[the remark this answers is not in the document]
joeduffy: D'oh yes, good catch. Will fix.

sdk/nodejs/policy/policy.ts
```diff
@@ -163,6 +163,14 @@ export interface PolicyConfigSchema {
  */
 export type Policies = (ResourceValidationPolicy | StackValidationPolicy)[];
 
+/*
+ * ResourceRemediation is a callback responsible for remediating a resource policy violation. It can either return
+ * new resource properties to be substituted for the old ones, or undefined if no remediation took place.
```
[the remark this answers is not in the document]
joeduffy: Yeah, I suppose there _might_ be some way to allow options to be changed if we move the remediation step to much earlier in the process. Maybe as soon as a `RegisterResource` call is received? Even there, though, there are options like parent that I fear could cause issues if the state isn't reflected back into the client process. I am happy to future proof it if we think there's ever a path to supporting this, however, it does seem unlikely. (Famous last words...)

sdk/nodejs/policy/policy.ts
```diff
@@ -385,6 +439,36 @@ export function validateResourceOfType<TResource extends Resource, TArgs>(
...
+ */
+export type TypedResourceValidationRemediation<TProps> =
+    (props: TProps, args: ResourceValidationArgs, reportViolation: ReportViolation) =>
+        Promise<Record<string, any>> | Record<string, any> | Promise<void> | void | undefined;
+
+/**
+ * A helper function for the pattern where a single function wants to be able to remediate *and*
+ * validate depending on how it is called.
+ */
+export function validateRemediateResourceOfType<TResource extends Resource, TArgs>(
```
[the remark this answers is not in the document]
joeduffy: Good idea, will do.

joeduffy:
> What are we going to do about secrets? 

That is one of the few things not done just yet, but that I'll be completing over the weekend. I called out the approach in [the design doc](https://docs.google.com/document/d/1ju4NmsW9D36CFiiiIzpreq_kVl0BwKWHVqgtawm4U1M/edit?usp=sharing) -- see "Dealing with Secrets". The idea was to automatically re-apply secret-ness to anything that was a secret prior to calling the remediation. This feels like it is the least surprising and does the right thing by default.

I had not considered a new proxy that can keep track of secrets ... That may be another approach to accomplish the same, although we'd need a change at the gRPC layer to send the secrets over to the analyzer instead of plaintexting them.

The nice thing about the engine doing this is a remediation can in fact just remediate a property like usual, even if it is secret, and the engine will just re-apply the secret-ness back on the engine side of the call.

That said, I do like your idea of being about to use `pulumi.secret` / `Output.secret` _in addition_ to whatever we do here.

What do you think about this approach?

sdk/nodejs/policy/policy.ts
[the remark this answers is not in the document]
justinvp: Let's keep it how you have it currently.

justinvp:
> What do you think about this approach?

Yeah, I had that in mind as well. I think it's reasonable. One thing I'm curious about: Will deeply nested secrets be preserved or will the entire top-level property be made a secret when re-applying secret-ness back on the engine side of the call if the property had originally contained a secret? The reason I ask is because users are often frustrated when a secret in a nested array or map ends up "tainting" the entire property as a secret, making diffs difficult to reason about. They don't want to see the whole property being a secret, they want the one secret value nested within a array/map to show as a secret, and not any other items that aren't secrets.

> although we'd need a change at the gRPC layer to send the secrets over to the analyzer instead of plaintexting them

Note: we already keep secrets when sending properties to the analyzer, and it looks like the current implementation in your PR does for `Remediate` as well:

https://github.com/pulumi/pulumi/blob/daaaaea339fc9675b1fec1d9893791362ade7da9/sdk/go/common/resource/plugin/analyzer_plugin.go#L176

https://github.com/pulumi/pulumi/pull/14080/files#diff-dd48d49dec6689ee06eaea36251a5881a54353583e2a6e9e88498f8fdf38c424R289-R290

They're stripped in the deserialization code in the policy SDK:

https://github.com/pulumi/pulumi-policy/blob/1f97d95fa3c91c56ff88c0e29d0da824bd22f597/sdk/nodejs/policy/deserialize.ts#L106-L107

https://github.com/pulumi/pulumi-policy/blob/1f97d95fa3c91c56ff88c0e29d0da824bd22f597/sdk/python/lib/pulumi_policy/deserialize.py#L57-L60

joeduffy:
The proxying approach worked quite well and it's easier to make it perfectly granular. I just pushed the changes. I explored just using `pulumi.secret` to represent the fact that properties are runtime secrets, which would have been nice because it would enable remediations to return `pulumi.secret`s explicitly to introduce new secrets into the state. Unfortunately, secrets are all wrapped up in the output machinery which entails promises, and this makes it impossible to promptly proxy values because even if we reached into its innards, we'd need to await something. It's possible we could do something dirty like monkey patch the `pulumi.secret` function to avoid all of that work in a way that remediations can recognize.

For now, I'll just go with the current approach as it at least preserves secrets. Feedback welcome!

sdk/nodejs/policy/server.ts
joeduffy: Indeed! I have added that as part of introducing secrets support.

sdk/nodejs/policy/tests/util.ts
```diff
@@ -56,11 +56,28 @@ export async function runResourcePolicy(resPolicy: policy.ResourceValidationPoli
...
     for (const validation of validations) {
-        await Promise.resolve(validation(args, report));
+        if (validation) {
+            await Promise.resolve(validation(args, report));
+        }
     }
     return violations;
 }
 
+export type PolicyRemediation = Record<string, any> | undefined;
```
justinvp, line 66:
```suggestion
/** @internal */
export type PolicyRemediation = Record<string, any> | undefined;
```

sdk/python/lib/pulumi_policy/secret.py
```diff
@@ -0,0 +1,92 @@
+# Copyright 2016-2020, Pulumi Corporation.
```
justinvp, line 1:
```suggestion
# Copyright 2016-2023, Pulumi Corporation.
```

tests/integration/resource_options/program/resource.ts
```diff
@@ -18,5 +18,8 @@ export class Provider implements pulumi.dynamic.ResourceProvider {
 export class Resource extends pulumi.dynamic.Resource {
     constructor(name: string, opts?: pulumi.CustomResourceOptions) {
         super(Provider.instance, name, {}, opts);
+        if (name === "aliased") {
+            console.log("OPTS = " + JSON.stringify(opts));
+        }
```
justinvp, line 23: Do we need to keep this? I assume it was to debug the alias thing per comments in the policy packs for this test.

justinvp (approved):
LGTM

PR #322, opened 2024-01-04
Switch to API Token for PyPi uploads

justinvp:
Username/Password authentication is no longer supported. This PR migrates to using API Tokens per https://pypi.org/help/#apitoken. We use `__token__` for the username and the org-wide `PYPI_API_TOKEN` secret for the password.

Reference:
- https://github.com/pulumi/pulumi/pull/15048
- https://github.com/pulumi/ci-mgmt/issues/751
- https://github.com/pulumi/ci-mgmt/pull/767


justinvp:
I'll need to fix the test failure due to change in `pulumi/pulumi`. Looks like we're missing a CRON worflow to run the tests on a daily basis.

Zaid-Ajaj (approved):

PR #381, opened 2025-06-27
Relax protobuf dependency constaint

Frassle:


Frassle:
> Guessing we've manually confirmed this will still work (or there are tests in this repo that confirm that)?

Seeing if it passes the repo tests

PR #397, opened 2025-09-04
[sdk/nodejs] Add support for new metadata fields

justinvp:
Add support for specifying additional metadata about the policy pack and policies within.

Example usage:

```typescript
new PolicyPack("initech-aws-policies", {
    displayName: "Initech AWS Policies",                // NEW
    description: "A detailed description...",           // NEW
    readme: `These policies check x, y, and z.
1. x - foo
2. y - bar
3. z - baz
`,                                                      // NEW
    provider: "AWS",                                    // NEW
    tags: ["aws"],                                      // NEW
    repository: "https://github.com/..."                // NEW
    policies: [{
        name: "s3-no-public-read",
        displayName: "No public-read S3 buckets",       // NEW
        description: "Prohibits setting...",
        severity: "medium",                             // NEW
        framework: {                                    // NEW
            name: "hitrust",                            // NEW
            version: "11.5",                            // NEW
            reference: "06.d Data Protection...",       // NEW
            specification: "Covered information...",    // NEW
        },
        tags: ["storage"],                              // NEW
        remediationSteps: "Change the bucket's acl...", // NEW
        url: "http://foo.bar/aws/s3-no-public",         // NEW
        validateResource: validateResourceOfType(aws.s3.Bucket, (bucket, args, reportViolation) => {
            if (bucket.acl === "public-read" || bucket.acl === "public-read-write") {
                reportViolation("You cannot set public-read or public-read-write on an S3 bucket.");
            }
        }),
    }],
});
```


sdk/nodejs/policy/policy.ts
```diff
@@ -35,6 +35,36 @@ export interface PolicyPackArgs {
      * override.
      */
     enforcementLevel?: EnforcementLevel;
+
+    /**
+     * A brief description of the policy pack. A description in PulumiPolicy.yaml will override this value.
+     */
+    description?: string;
```
justinvp, line 42: > A description in PulumiPolicy.yaml will override this value.

@danbiwer pointed out that we have a description already that can be specified in `PulumiPolicy.yaml`, though, it's not currently used anywhere. When creating a new policy pack via `pulumi policy new` a description will be emitted in the `PulumiPolicy.yaml` file from the policy template. (This is a historical accident with how `pulumi policy new` works based on how `pulumi new` works for projects).

> [!NOTE]
> The `PulumiPolicy.yaml` file's primary purpose is to indicate the runtime of the policy pack, e.g. `nodejs` vs `python`. It also serves as the place where the version is specified for Python policy packs. The version for Node.js policy packs is specified in `package.json`, but a value in `PulumiPolicy.yaml` will override it.

So what do we do about the description? I don't really want users to have to specify the description in the file and everything else in code. I also don't want users to have to specify all the policy pack fields in the file. We already have the precent of specifying existing fields for policies in code, and it feels natural to have the fields for the policy pack defined in the code similarly. (This is "policy as code", after all). That way, you're declaring all the metadata about the policies *and* policy pack in one place: the code[^1].

So if we want to support specifying it in code, what do we do about the description in the file?

1. Ignore description in the file
2. Use description in the file if not specified in code and vice-versa
    a. If specified in both, file overrides the code
    b. If specified in both, code overrides the file

I've been leaning towards (2a). Do folks feel strongly we should do something else?

Either way, I don't like that there's multiple ways to do things, so if our preference is the code, we should update our policy templates and `pulumi policy new` to stop emitting the description in the file.

sdk/nodejs/policy/policy.ts
```diff
@@ -84,6 +114,11 @@ export class PolicyPack {
  */
 export type EnforcementLevel = "advisory" | "mandatory" | "remediate" | "disabled";
 
+/**
+ * Indicates the severity of a policy.
+ */
+export type Severity = "low" | "medium" | "high" | "critical";
```
justinvp, line 120: For policies that don't set a severity, what comes back from `GetAnalyzerInfo` rpc call and included in the publish API request? Could either be empty or unspecified or we could have it default to some value, probably `"medium"`.

sdk/nodejs/policy/policy.ts
```diff
@@ -139,6 +174,56 @@ export interface Policy {
...
+
+    /**
+     * The compliance framework that this policy belongs to.
+     */
+    framework?: PolicyComplianceFramework;
+
+    /**
+     * Tags associated with the policy.
+     */
+    tags?: string[];
```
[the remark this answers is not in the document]
danbiwer: tags makes sense to me, I just took "topics" from the compliance policy manager

danbiwer:
Looks good to me, the fields mostly match what we've been tracking for the new policies (with the exception of readme, remediationSteps, and repository, but we can go back and set those)

I realize this was missed in the other threads, but came up in some conversations.  Do we want to support a "link" field to supporting links for a policy?  If there's no need for that in the UI, maybe we can just add that in a future update.

sdk/nodejs/policy/policy.ts
AnuiDolphin: This is one of those things that we will be wrong no matter what we choose, but I would default to low and require users to elevate the violation themselves.

arunkumar611: Any concerns with keeping it as "empty" or "Not specified"? 

It makes it explict to users that we haven't made a determination if the pack is Pulumi-authored. If the pack is custom, then it is clear to them that they haven't specified it.

sdk/nodejs/policy/policy.ts
justinvp: Going with 2b

sdk/nodejs/policy/policy.ts
justinvp: At the protobuf layer, there is an "unspecified" value and at the API layer between the CLI and the service, if the severity isn't specified, it will be omitted from the request. So you're choice on how you want it to show up in the UI, whether it unspecified shows up as low or unspecified.

PR #413, opened 2025-09-17
[sdk/nodejs] Replace `yarn` with `bun`

justinvp:
Use `bun install` rather than `yarn install`.

This speeds up the integration tests in CI from ~9-10min to ~6.5min.


tgummerer (approved):
:runner:

PR #424, opened 2025-10-20
Add overload for reportViolation to allow passing an object argument

fallimic:
This PR adds support for passing an object to the reportViolation method, while keeping the overload private for now.

### DIFF
--- a/sdk/nodejs/policy/policy.ts
+++ b/sdk/nodejs/policy/policy.ts
@@ -798,9 +798,26 @@ export function validateStackResourcesOfType<TResource extends Resource>(
 }
 
 /**
+ * The set of arguments used to report a policy violation.
+ *
+ * @internal
+ */
+export interface ReportViolationArgs {
+    message: string;
+    urn?: string;
+    name?: string;
+    description?: string;
+    enforcementLevel?: EnforcementLevel;
+}
+
+/**
  * ReportViolation is the callback signature used to report policy violations.
  */
-export type ReportViolation = (message: string, urn?: string) => void;
+export type ReportViolation  = {
+    (message: string, urn?: string): void;
+    // Future support for passing args object
+    // (args: ReportViolationArgs): void;
+};
 
 /**
  * Secret allows values to be marked as sensitive, such that the Pulumi engine will encrypt them
--- a/sdk/nodejs/policy/server.ts
+++ b/sdk/nodejs/policy/server.ts
@@ -35,6 +35,7 @@ import {
     PolicyResource,
     PolicyResourceOptions,
     ReportViolation,
+    ReportViolationArgs,
     ResourceValidationArgs,
     StackValidationArgs,
     getPulumiType,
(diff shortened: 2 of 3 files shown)


sdk/nodejs/policy/policy.ts
```diff
@@ -797,10 +797,25 @@ export function validateStackResourcesOfType<TResource extends Resource>(
...
 
+/**
+ * The set of arguments used to report a policy violation.
+ */
+export interface ReportViolationArgs {
+    urn: string;
+    name: string;
+    description: string;
+    enforcementLevel?: EnforcementLevel;
+    severity?: Severity;
```
justinvp: Let's remove this for now. I'll add it back after this PR when I plumb severity through over gRPC (working on that now).

sdk/nodejs/policy/tests/server.spec.ts
```diff
@@ -901,6 +902,57 @@ describe("makeRemediateRpcFun", () => {
             assert.equal(response!.getNotApplicableList()[0].getReason(), "Policy does not implement remediateResource");
         }));
     });
+
+    describe("ReportViolation with object argument", () => {
```
justinvp: I think this should be within the top-level `describe("makeAnalyzeRpcFun", () => {`, so right after line 476.

justinvp (commented):
Looks pretty good! Some comments.

sdk/nodejs/policy/server.ts
```diff
@@ -819,3 +804,45 @@ function pushNotApplicable(
...
+    return (message, urn?) => {
+        if (typeof message === "object") {
+            const policyInfo: ReportViolationArgs = message;
+            ds.push({
+                policyName: policyInfo.name || policyName,
+                policyPackName,
+                policyPackVersion,
+                message: policyInfo.message,
+                urn: policyInfo.urn,
+                description: policyInfo.description || policyInfo.message,
```
justinvp: I'm wondering if this should be:

```suggestion
                description: policyInfo.description || description,
```

justinvp (approved):
A couple other questions, but LGTM otherwise.

sdk/nodejs/policy/server.ts
```diff
@@ -819,3 +804,45 @@ function pushNotApplicable(
...
+    ds: Diagnostic[],
+): ReportViolation {
+    return (message, urn?) => {
+        if (typeof message === "object") {
+            const policyInfo: ReportViolationArgs = message;
+            ds.push({
+                policyName: policyInfo.name || policyName,
+                policyPackName,
+                policyPackVersion,
+                message: policyInfo.message,
```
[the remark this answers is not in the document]
fallimic: I think that's actually confusing behavior (and also undocumented afaik). And in the case of a dynamic policy violation, it ends up with the user being forced to set the description or the message to be empty. 

Essentially in my policy I might report a dynamic violation and want to pass a single string (message) that will be used and not join with the original description. Having to explicitly set description to "" feels inconvenient in this case. That's why I opted to avoid it for the object args.

PR #442, opened 2026-05-18
Fix Proxy invariant violation on Array.filter/map in unknownCheckingProxy

jkodroff:
## Summary

Fixes a V8 Proxy invariant `TypeError` thrown when policy code calls `Array.prototype.filter` (or any species-creating method: `.map`, `.slice`, `.concat`, etc.) on a resource field reached through `unknownCheckingProxy`. Reported by a customer in [pulumi/customer-support#2779](https://github.com/pulumi/customer-support/issues/2779).

The exact error customers see:

```
TypeError: 'get' on proxy: property 'prototype' is a read-only and
non-configurable data property on the proxy target but the proxy did
not return its actual value (expected '[object Array]' but got '[object Array]')
    at Proxy.filter (<anonymous>)
```

## Root cause

`Array.prototype.filter` internally invokes `ArraySpeciesCreate`, which:
1. Reads `.constructor` on the proxied array → gets a Proxy wrapping `Array`.
2. Looks up `[Symbol.species]` → gets a Proxy wrapping `Array` again.
3. Calls `new ProxiedArray(0)` → triggers V8's internal lookup of `.prototype` on the proxied constructor.

`Array.prototype` is a `{configurable: false, writable: false}` own data property. The Proxy invariant requires the `get` trap return the exact same value for such properties — but `proxyHelper` was wrapping it in another proxy, violating the invariant.

## Fix

In `proxyHelper`'s `get` trap, check the property descriptor before wrapping. For non-configurable + non-writable own data properties, return the raw value. These properties (`Array.prototype`, `Function.prototype`, etc.) are never Pulumi resource fields and never carry unknown sentinels, so bypassing the wrap is safe.

```diff
+ const desc = Object.getOwnPropertyDescriptor(obj, prop);
+ if (desc && "value" in desc && desc.configurable === false && desc.writable === false) {
+     return field;
+ }
  return proxyHelper(field, newProps);
```

## Re-export

`unknownCheckingProxy` and `UnknownValueError` are now re-exported from the package entry point so downstream policy-pack authors can use them in test helpers to mirror real runtime semantics. This is needed by [pulumi/policy-packs-internal#TBD](https://github.com/pulumi/policy-packs-internal), which routes its test args through the same proxy and surfaces a whole class of latent bugs that previously only manifested in the field.

## Commits

1. `Add failing tests for Array.filter/map on proxied resource args` — three regression tests that pre-fix throw the exact V8 TypeError customers see, post-fix pass.
2. `Fix Proxy invariant violation on Array.filter/map in unknownCheckingProxy` — the descriptor bypass + the re-export.

## Test plan

- [x] All three new tests in `proxy.spec.ts` throw the V8 TypeError pre-fix and pass post-fix
- [x] Full SDK test suite: 67 passing post-fix (2 pre-existing unrelated failures in `deserialize.spec.ts` — same failures occur on `main`)
- [x] End-to-end repro: built the CIS Azure pack against this branch via yalc, ran `pulumi preview` against a `DiagnosticSetting` matching Veeam's shape — clean policy advisory instead of crash
- [ ] Reviewer: confirm no regressions in downstream consumers that may depend on specific proxy behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk/nodejs/policy/proxy.ts
+++ b/sdk/nodejs/policy/proxy.ts
@@ -61,6 +61,17 @@ function proxyHelper<T>(toProxy: any, propsAcc: (keyof T)[]): any {
             if (isUnknown(field)) {
                 throw new UnknownValueError<T>(field, newProps);
             }
+            // Honor the V8 Proxy invariant: for an own data property that is both
+            // non-configurable and non-writable, the get trap must return the exact
+            // same value as the target. Wrapping such a property in another proxy
+            // (as proxyHelper does for everything else) triggers a TypeError. This
+            // matters because Array.prototype.filter (and other species-creating
+            // methods) internally accesses `.prototype` on the proxied Array
+            // constructor — a property that is {configurable: false, writable: false}.
+            const desc = Object.getOwnPropertyDescriptor(obj, prop);
+            if (desc && "value" in desc && desc.configurable === false && desc.writable === false) {
+                return field;
+            }
             return proxyHelper(field, newProps);
         },
     });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## HEAD (Unreleased)
 
+- Node.js: Fix `TypeError: 'get' on proxy: property 'prototype' is a read-only and non-configurable data property…` thrown when a policy calls `Array.prototype.filter` (or `.map`, `.slice`, `.concat`, or any other method that uses `ArraySpeciesCreate`) on a resource field reached through `unknownCheckingProxy`. The proxy's `get` trap now honors V8's invariant for non-configurable + non-writable own data properties, scoped narrowly enough to leave unknown-value detection during preview fully intact.
+  (https://github.com/pulumi/pulumi-policy/pull/442).
+
+- Node.js: Re-export `unknownCheckingProxy` and `UnknownValueError` from the package entry point so policy-pack test helpers can mirror real runtime semantics.
+  (https://github.com/pulumi/pulumi-policy/pull/442).
+
 ---
 
 ## 1.20.0 (2025-10-23)
--- a/sdk/nodejs/policy/index.ts
+++ b/sdk/nodejs/policy/index.ts
@@ -13,3 +13,4 @@
 // limitations under the License.
 
 export * from "./policy";
+export { unknownCheckingProxy, UnknownValueError } from "./proxy";
--- a/sdk/nodejs/policy/tests/proxy.spec.ts
+++ b/sdk/nodejs/policy/tests/proxy.spec.ts
@@ -190,4 +190,48 @@ describe("proxy", () => {
             .map((l: any) => l.category);
         assert.deepStrictEqual(enabledCategories, ["AuditLogs", "GatewayLogs"]);
     });
+
+    it("throws when .filter encounters an unknown array element", () => {
+        // Sibling to the existing forEach test: the descriptor bypass must not
+        // disable unknown-value detection during iteration. The first read of an
+        // unknown element should still throw UnknownValueError with the correct path.
+        assertThrowsUnknownValue(
+            () => {
+                const props: any = unknownCheckingProxy({ foo: [true, unknownBooleanValue, false] });
+                return props.foo.filter((x: any) => x);
+            },
+            unknownBooleanValue,
+            ["foo", "1"],
+        );
+    });
+
+    it("throws when .map encounters an unknown array element", () => {
+        assertThrowsUnknownValue(
+            () => {
+                const props: any = unknownCheckingProxy({ foo: [true, unknownBooleanValue, false] });
+                return props.foo.map((x: any) => x);
+            },
+            unknownBooleanValue,
+            ["foo", "1"],
+        );
+    });
+
+    it("throws when .filter callback reads an unknown nested property", () => {
+        // Array elements themselves are known objects, but a field inside one is unknown.
+        // The callback's `.enabled` access on the proxied object must still throw.
+        assertThrowsUnknownValue(
+            () => {
+                const props: any = unknownCheckingProxy({
+                    logs: [
+                        { enabled: true },
+                        { enabled: unknownBooleanValue },
+                        { enabled: false },
+                    ],
+                });
+                return props.logs.filter((l: any) => l.enabled);
+            },
+            unknownBooleanValue,
+            ["logs", "1", "enabled"],
+        );
+    });
 });
--- a/sdk/nodejs/policy/tests/proxy.spec.ts
+++ b/sdk/nodejs/policy/tests/proxy.spec.ts
@@ -154,4 +154,40 @@ describe("proxy", () => {
             ["foo", "1"],
         );
     });
+
+    it("supports Array.prototype.filter on proxied arrays", () => {
+        // Regression test: Array.prototype.filter internally uses ArraySpeciesCreate,
+        // which looks up `.constructor[Symbol.species]` and then accesses `.prototype`
+        // on the resulting (proxied) constructor. `Array.prototype` is a non-configurable,
+        // non-writable data property; the Proxy invariant requires the get trap to
+        // return the exact same value. Without the descriptor bypass in proxyHelper,
+        // this throws a V8 TypeError.
+        const props: any = unknownCheckingProxy({
+            logs: [{ enabled: true }, { enabled: false }, { enabled: true }],
+        });
+        const enabled = props.logs.filter((l: any) => l.enabled);
+        assert.strictEqual(enabled.length, 2);
+    });
+
+    it("supports Array.prototype.map on proxied arrays", () => {
+        const props: any = unknownCheckingProxy({
+            logs: [{ enabled: true }, { enabled: false }],
+        });
+        const flags = props.logs.map((l: any) => l.enabled);
+        assert.deepStrictEqual(flags, [true, false]);
+    });
+
+    it("supports chained .filter().map() on proxied arrays", () => {
+        const props: any = unknownCheckingProxy({
+            logs: [
+                { enabled: true, category: "AuditLogs" },
+                { enabled: false, category: "RequestLogs" },
+                { enabled: true, category: "GatewayLogs" },
+            ],
+        });
+        const enabledCategories = props.logs
+            .filter((l: any) => l.enabled)
+            .map((l: any) => l.category);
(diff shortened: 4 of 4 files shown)


Frassle (commented):
What if there are unknowns inside the array? Can we test how that behaves?

jkodroff:
Good question — added three tests in bd54ed8:

1. `.filter` over `[true, unknownBooleanValue, false]` throws `UnknownValueError` at path `["foo", "1"]`.
2. Same for `.map`.
3. `.filter(l => l.enabled)` over an array where one element has `enabled: unknownBooleanValue` throws at the deeper path `["logs", "1", "enabled"]`.

The descriptor bypass is scoped to non-configurable + non-writable data properties — sentinels are plain string values that never match that descriptor shape, so the unknown check runs as normal during iteration. These tests prove that.

jkodroff:
@Frassle That's Claude talking above. As to whether that is acceptable behavior, I do not know because this is beyond my depth in terms of understanding the policy engine.

BUT, Claude's explanation seems plausible at least:

Good catch to push on. Confirmed by reading server.ts:340-360: UnknownValueError is the designed signal that a policy can't be evaluated during preview. The server catches it, downgrades enforcement to "advisory", and emits this diagnostic:

  can't run policy '<name>' from policy pack '<pack>' during preview: <type> value at .<path> can't be known during preview

So the throw isn't a crash — it's how the SDK tells the engine "defer this policy until update, when the value will be known." Three additional reasons it's the right behavior for .filter/.map:

1. Consistency with existing API. The SDK already throws on .forEach with an unknown element (the pre-existing test at proxy.spec.ts:145). My tests just establish that .filter/.map behave the same way.

2. Filter semantics demand it. If [known, unknown, known] and the predicate is x => x > 5, the result can't be computed — you'd have to guess whether to include the unknown. Producing a "best-effort" filtered array would let a policy report a false positive (or false negative) violation against the customer's stack. Throwing forces the deferral.

3. The throw is path-tagged. The error message tells the user exactly which property was unknown (["logs", "1", "enabled"]), so policy authors can write narrower checks (e.g., gate on if (l.enabled === false) after first checking for unknownness) if they need to evaluate during preview.

If the team wanted different semantics (e.g., skip unknown elements silently), that'd be a separate design change — but it would be a behavior change for .forEach too, not just my new tests. My tests are documenting the current contract, which the descriptor-bypass fix correctly preserves.

sdk/nodejs/policy/proxy.ts
```diff
@@ -61,6 +61,17 @@ function proxyHelper<T>(toProxy: any, propsAcc: (keyof T)[]): any {
...
+            // Honor the V8 Proxy invariant: for an own data property that is both
+            // non-configurable and non-writable, the get trap must return the exact
+            // same value as the target. Wrapping such a property in another proxy
+            // (as proxyHelper does for everything else) triggers a TypeError. This
+            // matters because Array.prototype.filter (and other species-creating
+            // methods) internally accesses `.prototype` on the proxied Array
+            // constructor — a property that is {configurable: false, writable: false}.
+            const desc = Object.getOwnPropertyDescriptor(obj, prop);
+            if (desc && "value" in desc && desc.configurable === false && desc.writable === false) {
+                return field;
```
julienp, line 73: Presumably the proxy thing is there for a reason though.

Looking at history, this is there to help with policy running during preview. Are we sure we're not breaking anything here?

jkodroff:
The draft PR for the CIS Azure policy pack has some additional context that's worth reading: https://github.com/pulumi/policy-packs-internal/pull/180

And here's a sample resource that will reproduce the bug (against the CIS Azure pack):

```typescript
new azure.monitor.DiagnosticSetting("diag-firewall-metrics", {
    resourceUri: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/repro-rg/providers/Microsoft.Network/azureFirewalls/repro-fw",
    eventHubName: "logs-eventhub",
    eventHubAuthorizationRuleId: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/repro-rg/providers/Microsoft.EventHub/namespaces/repro-ns/authorizationRules/RootManageSharedAccessKey",
    logs: [
        {
            category: "AzureFirewallApplicationRule",
            enabled: true,
            retentionPolicy: { enabled: true, days: 365 },
        },
        {
            category: "AzureFirewallNetworkRule",
            enabled: true,
            retentionPolicy: { enabled: true, days: 365 },
        },
    ],
    metrics: [
        { category: "AllMetrics", enabled: true },
    ],
});
```

jkodroff:
*Written by Claude on Josh's behalf — capturing the explanation from our review-thread back-and-forth in one place so future readers don't have to piece it together.*

## Why this fix is narrow and safe

### Where the bug comes from

`Array.prototype.filter` (and `.map`, `.slice`, `.concat`, `.flat`, `.flatMap`, `.splice`, `.toReversed`, `.toSorted`, `.toSpliced`, `.with`, plus the TypedArray equivalents, `String.prototype.split`/`.match`, `RegExp.prototype.exec`, and `Promise.all`/`.allSettled`/`.race`) all use the [`ArraySpeciesCreate`](https://tc39.es/ecma262/#sec-arrayspeciescreate) abstract operation to decide what kind of object to return. The relevant steps:

```js
const C = this.constructor;
const S = C[Symbol.species] ?? C;
return new S(length);   // ← V8 reads .prototype on S here
```

When `this` is `unknownCheckingProxy(someArray)`:

1. `this.constructor` goes through the proxy's `get` trap and returns `proxyHelper(Array)` — a Proxy wrapping the `Array` constructor function.
2. `[Symbol.species]` on that returns `proxyHelper(Array)` again (since `Array[Symbol.species] === Array`).
3. `new <proxiedArray>(0)` causes V8 to read `prototype` on the proxied `Array`. The descriptor on the underlying `Array` is `{value: <Array.prototype>, writable: false, configurable: false}`.

ECMA-262 §10.5.8 — the `[[Get]]` internal method for Proxy — requires that **for an own data property with `writable: false` and `configurable: false`, the trap's return value `SameValue` the target's value**. Our trap returned `proxyHelper(Array.prototype)` (a wrapped Proxy) instead of `Array.prototype` itself. V8 detected the violation and threw the `TypeError` the customer saw.

### What the fix changes

Exactly one new branch in `proxyHelper`'s `get` trap, *after* the unknown-value check:

```ts
const desc = Object.getOwnPropertyDescriptor(obj, prop);
if (desc && "value" in desc && desc.configurable === false && desc.writable === false) {
    return field;
}
```

The `"value" in desc` guard restricts this to **data properties** (descriptors with a literal `value`, as opposed to accessor properties that use `get`/`set` functions). V8's invariant only applies to data properties — accessor properties are checked differently — so scoping the bypass to data props keeps it as narrow as the spec actually requires.

### Where the bypass actually fires

Walking through every kind of object `unknownCheckingProxy` can wrap:

| Target | Property | Descriptor | Bypass? |
|---|---|---|---|
| Plain object `{name: "web"}` | any user prop | `{configurable: true, writable: true}` | no |
| Array `[1, 2, 3]` | `length` | `{configurable: false, writable: **true**}` | no (writable) |
| Array `[1, 2, 3]` | `"0"`, `"1"`, … | `{configurable: true, writable: true}` | no |
| Built-in ctor (`Array`, `Object`, `Function`, `Map`, …) reached via `.constructor` | **`prototype`** | `{configurable: false, writable: false}` | **YES** |
| Built-in ctor | `name`, `length` | `{configurable: true, writable: false}` | no (configurable) |

In practice the bypass fires in exactly one situation: when V8 walks through a proxied built-in constructor function to read its `.prototype`. It cannot match a Pulumi resource property because `deserializeProperties` produces plain JS objects via normal assignment / `Object.assign` / spread — those create `{configurable: true, writable: true}` slots. The SDK never calls `Object.defineProperty(obj, key, {configurable: false, writable: false})` anywhere in the policy-server code path (grep confirms).

### What's preserved

Unknown-value detection. Sentinel strings like `"04da6b54-80e4-46f7-96ec-b56ff0331ba9"` (unknown string) live at ordinary `{configurable: true, writable: true}` slots, so the descriptor check passes them through to the existing `isUnknown(field)` branch and they still throw `UnknownValueError` with the right path. The three tests added in bd54ed8 verify this directly through the new code paths:

- `.filter` over `[true, unknownBooleanValue, false]` throws at `["foo", "1"]`
- `.map` over the same throws at `["foo", "1"]`
- `.filter(l => l.enabled)` where one element has `enabled: unknownBooleanValue` throws at `["logs", "1", "enabled"]`

### Why throwing on unknowns is the right behavior

Picked up from the inline reply: `server.ts:340–360` catches `UnknownValueError` from a policy and downgrades that policy to `"advisory"` with the diagnostic `"can't run policy <X> from policy pack <Y> during preview: <type> value at .<path> can't be known during preview"`. The throw isn't a crash — it's the SDK's API for telling the engine "defer this policy until update, when the value will be known." The pre-existing `.forEach` test at `proxy.spec.ts:145` establishes this contract for iteration; the new tests just extend it to `.filter`/`.map` (which were previously unreachable because of the very TypeError this PR fixes).

Producing a "best-effort" filter result on partial data would let policies report false-positive or false-negative violations against customers' real infrastructure. Deferring is correct.

### What does NOT change

- Direct property access on resource props (`vm.publicIp`, `args.props.logs`, etc.) — bypass doesn't fire, unknown detection works exactly as before.
- `for...of`, `forEach`, manual indexing — bypass doesn't fire on user data; throws on unknowns as before.
- Mutation of `Array.prototype` from inside a policy — already possible before this PR. The SDK's proxy has no `set` trap, so writes always passed through to the real prototype. This change only affects the *read* path to match.

### Net effect

The proxy continues doing its one job (catching unknown sentinels during preview). It stops doing the thing it was never designed to do (interfering with V8's internal use of built-in prototypes). The customer-blocking crash that prompted this PR (pulumi/customer-support#2779) goes away, and the same fix transparently unblocks any other policy that uses array methods returning new arrays.

jkodroff:
I learned a lot about the policy engine from writing this PR...

sdk/nodejs/policy/proxy.ts
jkodroff: See https://github.com/pulumi/pulumi-policy/pull/442#issuecomment-4490400743

I'm leaning on Claude of course, but I did press him to really explain how this stuff works and why this should be safe.

Frassle (approved):
I think this is fine, the added tests show this behaving sensibly for nested unknowns which is what I was worried about.

PR #459, opened 2026-08-26
python: return exceptions correctly

### LINKED ISSUE #16741: Stack trace coming from Policy Pack should be displayed.
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

A customer reports that Pulumi signals an exception happened when running Pulumi with a policy pack, but the stacktrace is not shown. To find out where the problem lies, Pulumi should show this.

```
...
Diagnostics:
  pulumi:pulumi:Stack (<customer-stack>):
    Using environment test
    Using <masked> self service base url <masked>

  aws.iam.Policy (<customer-policy-pack-name>)
    error: failed to run policy: Exception calling application: 'Resource'
```

The above are the last lines of the current output.

(Working on reproducible case. Will update issue when done.)

### Affected area/feature

Pulumi CLI

tgummerer:
Return exceptions via the appropriate grpc mechanism instead of swallowing them.

Fixes pulumi/pulumi#16741

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## HEAD (Unreleased)
 
 - Python: Support protobuf v6.x
+- Python: Return exceptions correctly
 
 ## 1.21.0 (2026-05-20)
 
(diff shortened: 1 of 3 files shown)


Frassle (approved):
We should check if the provider SDK in pu/pu does this correctly as well!

PR #460, opened 2026-09-13
feat(go): add a Go SDK for writing policy packs

### LINKED ISSUE #230: Support for writing policies in Go

nghiadaulau:
Fixes #230

## Problem

Policy packs can be written in TypeScript/JavaScript and Python. There is no supported way to write them in Go (#230).

The Pulumi CLI can already run a Go policy pack in two ways:

- A directory whose `PulumiPolicy.yaml` sets `runtime: go`. `pulumi-language-go` compiles the pack and runs it through `RunPlugin`.
- A prebuilt binary passed to `--policy-pack`.

What is missing is a library that serves the Analyzer gRPC protocol with the same behavior as `@pulumi/policy` and `pulumi_policy`.

The Pulumi Go SDK ships an experimental package, `github.com/pulumi/pulumi/sdk/v3/go/pulumi/policyx`. In `sdk/v3` v3.262.0 it differs from the Node.js and Python SDKs, and its server type is unexported, so a wrapper cannot change that behavior:

- `NewPolicyPack` ignores its `enforcementLevel` argument.
- `Remediate` runs remediations at every enforcement level except `disabled`.
- Severity, not-applicable results, and initial config are never sent (`InitialConfig: nil, /* TODO */`).
- An unknown value during preview fails the whole `Analyze` call.
- `GetAnalyzerInfo` calls methods on a pack that only `ConfigureStack` creates, and `pulumi policy publish` does not call `ConfigureStack`.
- `DryRun` is never set. A `pulumi preview` smoke test printed `dryRun=false`.

## Approach

This PR adds a Go module, `github.com/pulumi/pulumi-policy/sdk/go`, with one package, `policy`. It implements `pulumirpc.AnalyzerServer` directly and uses the Pulumi Go SDK only for protocol types, property marshaling, and `rpcutil`. It requires `github.com/pulumi/pulumi/sdk/v3` v3.220.0 or later and Go 1.25.

```go
package main

import (
	"context"

	"github.com/pulumi/pulumi-policy/sdk/go/policy"
)

type bucketArgs struct {
	ACL string `pulumi:"acl"`
}

func main() {
	policy.Main("aws-go", policy.PolicyPackArgs{
		Policies: []policy.Policy{
			&policy.ResourceValidationPolicy{
				Name:             "s3-no-public-read",
				Description:      "Prohibits public read access on S3 buckets.",
				EnforcementLevel: policy.EnforcementLevelMandatory,
				ValidateResource: policy.ValidateResourceOfType("aws:s3/bucket:Bucket",
					func(_ context.Context, bucket bucketArgs, _ policy.ResourceValidationArgs,
						reportViolation policy.ReportViolation,
					) error {
						if bucket.ACL == "public-read" {
							reportViolation("You cannot set public-read on an S3 bucket.", "")
						}
						return nil
					}),
			},
		},
	})
}
```

### Behavior

| Topic | Go SDK behavior |
|---|---|
| Enforcement level | Configuration, then the policy's level, then the pack default, then `advisory`. `Analyze` and `AnalyzeStack` report `remediate` as `mandatory`. Same order as `server.ts` and `policy.py`. |
| Remediation | Runs only when the resolved level is `remediate`. Remediations run in order, and each one sees the previous result. |
| Not applicable | `return policy.NotApplicable(reason)`. `ValidateResourceOfType`, `RemediateResourceOfType`, `ValidateRemediateResourceOfType`, and `ValidateStackResourcesOfType` use the Node.js reason text: `Policy only applies to resources of type '<type>'`. |
| Unknown values | `DecodeProps` returns `*policy.UnknownValueError` for computed values. Calling an `As*` method on a computed `property.Value` is caught as well. Both produce the advisory diagnostic `can't run policy '<name>' from policy pack '<pack>@v<version>' during preview: ...`, as the other SDKs do. |
| Secrets | `Props` keep their secret flags. In a remediation result, a value that replaces data holding a secret stays secret. |
| Errors | Any other error, or a panic, fails the RPC with the policy name and pack in the message. Panics are recovered, so one policy cannot crash the pack process. |
| Metadata | `GetAnalyzerInfo` sends every field the Node.js SDK sends: display name, description, readme, provider, tags, repository, severity, compliance framework, remediation steps, URL, config schema, and initial config. A unit test fails when `AnalyzerInfo` or `PolicyInfo` gains a field the SDK does not map. |
| Stack policies | Resources link to their parent, dependencies, and property dependencies. Stack policies receive stack tags and config. |
| Concurrency | Config and stack tags are immutable snapshots behind `atomic.Pointer`. Each policy call receives cloned arguments. `ReportViolation` may be called from goroutines, and reports after the callback returns are dropped. |
| Cost | The engine calls `Analyze`, `AnalyzeStack`, and `Remediate` for every resource, so properties are decoded only when a policy is about to run. `AnalyzeStack` decodes resources on up to `GOMAXPROCS` goroutines. |

### Files

- `sdk/go/policy`: the SDK. `policy.go`, `errors.go`, and `oftype.go` hold the public API. `pack.go` holds validation and enforcement level resolution without protocol types. `analyzer.go` is the gRPC server. `decode.go` decodes `property.Map` into Go structs. `secrets.go` keeps remediated values secret. `parallel.go` is a bounded parallel loop. `serve.go` holds `Main` and `Run`.
- `sdk/go/Makefile`, root `Makefile`: `make lint` and `make test_fast` for the Go SDK, wired into `SUB_PROJECTS`.
- `.github/workflows/{main,release,run-acceptance-tests}.yml`: a `Lint Go` step in each lint job.
- `tests/integration`: `runPolicyScenarios` extracted from the existing closure, a `Check` hook on `policyTestScenario`, and `TestGoPolicyPack` with the `go_policy_pack` fixture.
- `design/03-go-sdk.md`, `sdk/go/README.md`, `README.md`, `CONTRIBUTING.md`, `CHANGELOG.md`: documentation.
- `.mise.toml`: Go 1.24 to 1.25, matching the `1.25.x` toolchain the workflows use.

## What this PR does

- [x] Go SDK with resource validation, resource remediation, and stack validation policies.
- [x] Typed helpers: `ValidateResourceOfType`, `RemediateResourceOfType`, `ValidateRemediateResourceOfType`, `ValidateStackResourcesOfType`.
- [x] `DecodeProps`: `pulumi` and `json` tags, embedded struct promotion, case-insensitive fallback, range-checked integers, secrets decoded to plaintext, `property.Value` passthrough.
- [x] Policy config through `args.GetConfig(&target)`, with `enforcementLevel` removed.
- [x] Pack metadata, policy metadata, config schema, and initial config in `GetAnalyzerInfo`.
- [x] `Handshake`, `ConfigureStack` (stack tags), `Configure`, `GetPluginInfo` (module version from build info), and `Cancel`.
- [x] Both CLI launch paths, tested end to end: `runtime: go` directory and prebuilt binary.
- [x] Unit tests, a fuzz target, an example test, and an integration test.
- [x] Make targets, CI lint steps, README, SDK README, design record, CONTRIBUTING release note, and a CHANGELOG entry.

## What this PR does not do

- **No `pulumi policy new` template for Go.** Templates live in a separate repository.
- **No engine access from policies.** Policy arguments do not expose invokes, stack config, project, stack, organization, or `DryRun`. The Node.js and Python policy arguments do not expose them either.
- **No Go packs for the existing fixtures.** Go coverage comes from the dedicated `go_policy_pack` fixture. The shared harness does not yet detect a `policy-pack-go/` directory next to `policy-pack/` and `policy-pack-python/`.
- **No release automation for the Go module tag.** `CONTRIBUTING.md` documents the manual `sdk/go/vX.Y.Z` tag. `release.yml` does not push it.
- **No fix for the engine behavior found during testing.** With Pulumi CLI v3.262.0, when a remediation forces the replacement of an existing resource, the new state keeps the program's original inputs. A Node.js pack gives the same result, so the problem is on the engine side and needs an issue in pulumi/pulumi. Details are under Verification.
- **No change to the unknown-value sentinel handling.** A program that sets a property to the literal unknown sentinel string makes a policy that reads it report an advisory diagnostic instead of a mandatory violation. The Node.js and Python SDKs behave the same way. A fix would use `DryRun` from `ConfigureStack` and should land in all three SDKs.
- **No direct protobuf-to-`property.Map` decoder.** Properties go through `resource.PropertyMap` first. A direct decoder would save 20 to 30 percent of decode time but would duplicate the secret, asset, and unknown encoding.
- **Not run locally:** `TestStackTags` (needs a Pulumi Cloud backend), Windows, and Node.js 20.

## Alternatives rejected

- **Wrap `policyx`.** The server is unexported, so enforcement level resolution, remediation gating, not-applicable results, severity, initial config, and the `pulumi policy publish` crash cannot be changed from outside. Its `go.mod` also requires Go 1.26.6.
- **Expose `resource.PropertyMap` to policies.** It is a mutable map. Sharing it across policies lets one policy change another's input, and copying it per call costs O(properties x policies) for each resource.
- **Version the Go module on its own `sdk/go/v0.x` line.** With `v1.22.0` and `sdk/go/v0.1.0` on the same commit, `pulumictl get version` reported `0.1.0` for the JavaScript and Python packages. With `sdk/go/v1.22.0` it reported `1.22.0`.

## Risk and rollback

- The `@pulumi/policy` and `pulumi_policy` packages are unchanged. `git diff` touches no file under `sdk/nodejs` or `sdk/python`.
- The shared change is the integration harness refactor. `runPolicyScenarios` runs the same commands in the same order, and the Node.js and Python integration results below match `main`.
- `.mise.toml` moves local development from Go 1.24 to 1.25. CI already uses `1.25.x`.
- The README labels the Go SDK as Preview. Once a `sdk/go/v1.x.y` tag is published, Go module rules require a `/v2` module path for breaking API changes, so the API should be reviewed before the first tag.
- Rollback: revert the PR. Nothing reaches Go users until a `sdk/go/vX.Y.Z` tag is pushed.

### Environment

Linux amd64, Go 1.27.0 toolchain (module `go 1.25.0`), Pulumi CLI v3.262.0, Node.js v24.20.0, Bun 1.4.2, Python 3.12.3, pipenv 2026.8.0, pulumictl v0.0.50.

### Go SDK: lint and unit tests

```
$ cd sdk/go && make lint test_fast
LINT:
go vet ./...
TEST FAST:
go test -race -count=1 -cover ./...
ok  	github.com/pulumi/pulumi-policy/sdk/go/policy	1.140s	coverage: 92.8% of statements
```

91 test functions (many table-driven), 1 example, and 1 fuzz target:

- Enforcement and remediation: `TestAnalyze_EnforcementLevelResolution` (10 cases), `TestRemediate_RunsOnlyAtRemediateLevel` (7 cases), `TestAnalyzeStack_EnforcementLevels`, `TestRemediate_MultiplePolicies_EachSeesPreviousResult`.
- Pack validation and metadata: `TestNewAnalyzer_InvalidArguments_ReturnsError` (18 cases), `TestGetAnalyzerInfo_WithoutStackConfiguration_ReturnsPackMetadata`, `TestGetAnalyzerInfo_FullyPopulatedPack_SetsEveryProtoField`, `TestGetAnalyzerInfo_InitialConfig_OmitsEmptyEntries`.
- Unknown values: `TestAnalyze_UnknownValueDuringPreview_ReportsAdvisoryAndContinues`, `TestAnalyze_ComputedValueReadWithAsMethod_ReportsAdvisory`, `TestAnalyze_OtherTypeAssertionPanic_ReturnsUnknownStatus`, `TestAnalyzeStack_UnknownValue_ReportsAdvisoryWithResourceURN`, `TestRemediate_UnknownValue_ReturnsDiagnostic`.
- Secrets: `TestPreserveSecrets` (9 cases), `TestRemediate_ReplacingSecretValues_KeepsThemSecret`, `TestDecodeProps_SecretNumberOutOfRange_ErrorOmitsValue`.
- Isolation and concurrency: `TestAnalyze_PolicyMutatesArgs_LaterPolicySeesOriginal`, `TestAnalyze_PolicyMutatesItsConfig_NextCallSeesOriginal`, `TestAnalyzeStack_PolicyMutatesResources_LaterPolicySeesOriginal`, `TestAnalyzeStack_PolicyMutatesTagsAndConfig_LaterPolicySeesOriginal`, `TestAnalyze_ReportAfterPolicyReturns_IsDropped`, `TestAnalyzeStack_ReportAfterPolicyReturns_IsDropped`, `TestAnalyze_ReportsFromGoroutines_AllCollected`, `TestAnalyze_ConcurrentWithReconfiguration_NoDataRace`.
- Lazy and parallel decoding: `TestAnalyze_NoPolicyRuns_SkipsDecodingProperties`, `TestAnalyzeStack_NoPolicyRuns_SkipsDecodingResources`, `TestRemediate_NoPolicyAtRemediateLevel_SkipsDecodingProperties`, `TestAnalyzeStack_ManyResources_DecodedInRequestOrder`, `TestAnalyzeStack_UndecodableResources_ReportsFirstInRequestOrder`, `TestForEachIndex_*`.
- Decoding: 17 `TestDecodeProps_*` tests covering scalars, number ranges, NaN and infinity, type mismatch paths, nested values, null, secrets, computed values, embedded structs, case-insensitive names, assets, and invalid targets.
- Server: `TestServe_WritesPortServesAnalyzerAndStopsOnCancel` (real gRPC client), `TestServe_PortWriteFails_StopsServerAndReturnsError`, `TestRun_PackAlreadyRunning_ReturnsError`, `TestModuleVersion`.
- `ExampleValidateResourceOfType_unitTest` shows how a pack author unit-tests a policy without the engine.

### Fuzzing

```
$ go test ./policy -run '^$' -fuzz=FuzzDecodeProps -fuzztime=30s
fuzz: elapsed: 30s, execs: 617800 (12040/sec), new interesting: 279 (total: 331)
PASS
```

### Vulnerability scan

```
$ go run golang.org/x/vuln/cmd/govulncheck@latest ./...
No vulnerabilities found.
Your code is affected by 0 vulnerabilities.
```

Before this result, grpc v1.71.1 made GO-2026-6061 reachable from the server transport. The module now requires grpc v1.82.1.

### Integration tests

```
$ cd tests/integration && go test . -v -count=1 -timeout 90m
```

| Test | Result |
|---|---|
| `TestGoPolicyPack` | PASS, 4 scenarios from the pack directory and 4 as a prebuilt binary |
| `TestConfig`, `TestEnforcementLevel`, `TestInvalidPolicy`, `TestParentDependencies`, `TestProvider`, `TestPythonDoesNotTryToInstallPlugin`, `TestRemoteComponent`, `TestResourceOptions`, `TestUnknownValues`, `TestValidatePythonResource`, `TestValidateResource`, `TestValidateStack` | PASS |
| `TestStackTags` | SKIP, needs `PULUMI_ACCESS_TOKEN`, `PULUMI_API`, and `PULUMI_TEST_OWNER` |
| `TestDeserialize`, `TestRuntimeData` | FAIL in the Node.js pack only, also on `main` (see below) |

`TestGoPolicyPack` scenarios:

1. First `pulumi up`: the config schema default (`minLength: 6`) produces an advisory violation for a 5-character string. `RandomPet.prefix` comes from an unknown output, so the preview reports `can't run policy 'pet-prefix-length' ... during preview`. A remediation replaces the secret `overrideSpecial` input, and `pulumi stack export --show-secrets` confirms the new value is `_` and still secret.
2. `--policy-pack-config` raises `min-length` to `mandatory` with `minLength: 10`, and the update fails with both violations.
3. A stack policy reads `maxResources: 1` from config and fails the update.
4. The remediation runs again on the existing resources, and the input stays `_` and secret.

**Failures present on `main`.** `TestDeserialize` and `TestRuntimeData` fail in their Node.js packs with `AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal` and `'allConfig()' not the expected value (via resource prop)`. Their Python packs pass. A clean clone of `af424fb` with the same toolchain fails with the same assertions. This environment runs Node.js 24 and `@pulumi/pulumi` 3.262.0, while `.mise.toml` pins Node.js 20. I did not confirm the cause on Node.js 20.

### Node.js and Python unit tests

- `sdk/python`: `make lint test_fast`, 51 tests, OK.
- `sdk/nodejs/policy`: 70 passing, 2 failing in `deserialize.spec.ts` (`marshals secrets correctly even when being proxied`, `round-trips proxied secrets correctly`). This PR does not change `sdk/nodejs`.

### Manual end-to-end checks

- **Launch paths.** With `-v=9` engine logging, both launch paths show `Handshake: success`, `StackConfigure: success`, and `Configure`, `Remediate`, and `Analyze` calls for every resource.
- **Remediation on replacement.** On a fresh stack, a Go pack and a Node.js pack both write `overrideSpecial: "_"` to state. When the stack already holds `overrideSpecial: "!@"`, both packs return the remediation (`#remediations=1`), the engine plans a replace on `overrideSpecial`, and the saved state still holds `"!@"`. Steps: `pulumi up` without policy config, then `pulumi up --policy-pack <pack> --policy-pack-config '{"override-special": {"enforcementLevel": "remediate"}}'`, then `pulumi stack export --show-secrets`.
- **Release tags.** In isolated clones, `pulumictl get version` reported `0.1.0` with `v1.22.0` and `sdk/go/v0.1.0` on one commit, and `1.22.0` with `v1.22.0` and `sdk/go/v1.22.0`.

### Review before opening

The change was reviewed for correctness, security, and performance before this PR was opened. Fixes that came out of the review:

- Decode errors no longer include property values, which could be secrets.
- Secret preservation marks arrays and values of a different kind as secret when the data they replace held a secret.
- Calling an `As*` method on a computed value produces the advisory diagnostic instead of failing the RPC.
- Properties are decoded only when a policy runs, and stack resources are decoded in parallel. On a benchmark with 10,000 resources of 12 KB, parallel decoding cut decode time from 3.58 s to 0.71 s on 16 cores.
- grpc v1.71.1 to v1.82.1 and golang.org/x/text v0.33.0 to v0.39.0 for known vulnerabilities.
- `Run` takes a `context.Context`, and timeouts use `time.Duration`.

## Notes for maintainers

- This PR comes from a fork, so the acceptance tests need a maintainer to comment `/run-acceptance-tests`.
- Releases need a `sdk/go/vX.Y.Z` tag with the same version as `vX.Y.Z`, as described in `CONTRIBUTING.md`.
- Commits are split by intent: harness refactor, SDK, build, CI, integration test, docs, and the changelog link.


pho-veteran:
I think the biggest impact here isn't just "Pulumi policies can now be written in Go." #230 has been open since 2020, and Go-heavy platform teams currently have to keep Node or Python in the toolchain just for policy packs. Being able to build and ship a single Go binary is a meaningful simplification.

What I'd be most careful about is maintenance. This is a fairly large direct implementation of the Analyzer protocol, so behavior could drift from the Node.js and Python SDKs over time. Preview feels right, and I'd spend most of the review time on the public API and parity tests before the first `sdk/go/v1` tag. Those will be much harder to change later than templates or release automation.

The unit and integration coverage in the PR already looks substantial. It'd be great to get `/run-acceptance-tests` triggered so the remaining discussion can focus on the API and overall direction.

nghiadaulau:
@Frassle @justinvp, could you take a look when you have time?

@Frassle, you mentioned on #230 that most of the Go policy work already exists in `policyx`, and that typed resource properties are the remaining gap. This PR takes a different route, so I would like to ask which direction you prefer before anyone spends time on a detailed review.

`policyx` on pulumi/pulumi `master` (the same files as v3.262.0) still differs from the Node.js and Python SDKs. Its server type is unexported, so a wrapper in this repository cannot change these:

- `NewPolicyPack` ignores its `enforcementLevel` argument.
- `Remediate` runs remediations at every enforcement level except `disabled`.
- Responses carry no severity, initial config (`InitialConfig: nil, /* TODO */`), or not-applicable results.
- An unknown value during preview fails the whole `Analyze` call.
- `GetAnalyzerInfo` calls methods on a pack that only `ConfigureStack` creates, and `pulumi policy publish` does not call `ConfigureStack`.

This PR also covers the typed properties you described: `ValidateResourceOfType[T]` and the other `...OfType` helpers decode resource properties into Go structs with `DecodeProps`.

Which do you prefer?

1. Review this PR as the Go SDK in pulumi-policy, next to the Node.js and Python SDKs.
2. I port the parity fixes, the typed helpers, and their tests into `policyx` as smaller PRs in pulumi/pulumi, and close this one.

Separately, while testing I found that when a remediation forces the replacement of an existing resource, Pulumi CLI v3.262.0 keeps the program's original inputs in the new state. A Node.js pack gives the same result, so it looks like engine behavior. The reproduction steps are in the PR description under "Manual end-to-end checks", and I can open an issue in pulumi/pulumi if that helps.

The acceptance tests need a maintainer to comment `/run-acceptance-tests`.

Frassle:
> Which do you prefer?

We have no intention to add Go policy support in this repo. Long term plans are that policy logic should live in the core SDKs, thus why policyx already exists. 

> NewPolicyPack ignores its enforcementLevel argument.
Remediate runs remediations at every enforcement level except disabled.
Responses carry no severity, initial config (InitialConfig: nil, /* TODO */), or not-applicable results.
An unknown value during preview fails the whole Analyze call.
GetAnalyzerInfo calls methods on a pack that only ConfigureStack creates, and pulumi policy publish does not call ConfigureStack.

Most of these sound like bugs in policyx. It would be ideal if possible to demonstrate that the feature works across all policy pack SDKs the same way by writing conformance tests for each one. Not everything fits into conformance testing though so just compare against how the other languages work for those bits.

> This PR also covers the typed properties you described: ValidateResourceOfType[T] and the other ...OfType helpers decode resource properties into Go structs with DecodeProps. 

We'll want to think carefully about what that should look like. Static types for policy has been a recurring discussion point internally about what's the 'right' way to expose them. As you'll have seen, just decoding to a plain struct hits a major issue in how to deal with `computed` values.

If you want to raise small PRs in pulumi/pulumi to fix the issues you've found with the server implementation so far we can take a look at them. Again if possible write conformance tests to show they're correct across all the languages at once.

nghiadaulau:
Thanks @Frassle, that makes sense. I'm closing this PR and moving the work to `policyx` in pulumi/pulumi:

- pulumi/pulumi#24646: serve analyzer info without stack configuration, so `pulumi policy publish` works for Go packs. Unit tests cover it, because the conformance runner always configures the stack.
- pulumi/pulumi#24647: run remediations only at the `remediate` enforcement level, with a new run in the `policy-remediate` conformance test that Go fails without the fix and Node.js and Python pass.
- pulumi/pulumi#24648: the gaps that need API changes (pack enforcement level, severity, not-applicable results, initial config, unknown values during preview), with a proposal for each.

I'll leave static types for policy to your internal discussion.

